PR #1, opened 2015-09-17
Add current members

tjconcept:
Would it make sense to move everything TSC here to avoid pollution of the node repository?

See nodejs/node#2925

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,22 @@ minutiae as the TSC delegates most of its responsibilities to other projects and
 Every Top Level Project not currently incubating can appoint someone to the TSC who they elect
 at their own discretion.
 
+### Current members
+
+* [bnoordhuis](https://github.com/bnoordhuis) - **Ben Noordhuis** &lt;info@bnoordhuis.nl&gt;
+* [chrisdickinson](https://github.com/chrisdickinson) - **Chris Dickinson** &lt;christopher.s.dickinson@gmail.com&gt;
+* [cjihrig](https://github.com/cjihrig) - **Colin Ihrig** &lt;cjihrig@gmail.com&gt;
+* [fishrock123](https://github.com/fishrock123) - **Jeremiah Senkpiel** &lt;fishrock123@rocketmail.com&gt;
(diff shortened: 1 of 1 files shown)


Fishrock123:
@tjconcept oops, looks like this was a while ago. :S

Those meeting don't belong here imo, they are mostly technical.

If you'd be willing to change this to just add the members here: https://github.com/nodejs/TSC#tsc-members I'll gladly take a look at the PR. :)

tjconcept:
Sure.

tjconcept:
Looks like you've semi-adopted a new commit message style since then - do you want me to prefix it with "doc:"?

Fishrock123:
@tjconcept That's git-style spill-over from nodejs/node. We wouldn't mind it, but you don't have to here.

tjconcept:
I assumed that. I will leave it then.

Fishrock123:
@tjconcept Ah sorry I didn't get to this sooner, we merged another thing with a description of TCS members, so this has conflicts? If you'd mind to rebase I can merge it asap. :)

tjconcept:
Yep, you want them below or above the new text?

Fishrock123:
Below is fine. :)

README.md
```diff
@@ -20,6 +20,20 @@ minutiae as the TSC delegates most of its responsibilities to other projects and
 Every Top Level Project not currently incubating can appoint someone to the TSC who they elect
 at their own discretion.
```
Fishrock123: Hmmm, maybe it would be best to have an `###` heading here saying something like "List of Current TSC Members"?

tjconcept: Yeah, that's the problem with having it below. It just seems a bit redundant to add another header?

If anything, I think "Current members" makes the most sense as the previous header already states "TSC".

williamkapke:
@tjconcept @Fishrock123 More thoughts/changes on this? I'd like to see it happen.

rvagg:
thanks @tjconcept, went with @williamkapke ✨'s

PR #4, opened 2015-10-06
minor formatting

Trott:
I know Charter changes need to go through a process, but this is just some minor formatting to make the Definitions section consistent. (Last definition is formatted differently than the first two in the current version of the doc.) So, uh, hopefully it can just sail in.


mikeal:
The actual document ratified by the board is a different text document, this is merely a copy that is here for convenience so we can make style changes easily. Content changes however would have to be approved by the board which is quite the process :)

PR #5, opened 2015-10-19
Travel fund for foundation members.

mikeal:
Based on feedback from the last board session I've drafted a fund for members to travel and promote the foundation.

This is based mostly on input from @danese on a fund the Apache Foundation had early on and on conversations with @indexzero and @rvagg

### DIFF
--- a/Member-Travel-Fund.md
+++ b/Member-Travel-Fund.md
@@ -19,7 +19,7 @@ New slide decks are added to the list of approved decks on an ongoing basis.
  the list above.
 * Slide deck must be approved by the TSC.
 
-Any member can apply to give an approved talk. Request may take up to a month
+Any member can apply to give an approved talk. Each request may take up to a month
 for the TSC to approve or decline.
 
 * Log an issue in this repository.
@@ -27,13 +27,18 @@ for the TSC to approve or decline.
  * Include the location of the event and where you will be traveling from.
  * Include the presentation you intend to give.
  * Include the size of the stipend you wish to receive.
-  * Stipends are anywhere from $100 to $1,500 USD and vary with travel distance.
+  * Reimbursement stipends are anywhere from $100 to $1,500 USD and vary with travel distance.
 
 The TSC will consider each proposal in its next regular meeting. While it is
 ultimately at the TSC's discretion who to approve the following considerations
 are made.
 
-* Impact: Preference given to underserved communities.
+* Impact: Preference given to underserved communities. More specifically,
+within a subject the foundation trying to promote awareness about (e.g. general
+knowledge about the foundation) there are people, communities and geographies
+where that knowledge is not very wide spread. This is where preference will
+be given. For instance, an event in San Francisco would be less attractive
+than an event of the same size in Kansas City.
 * Cost: The larger the stipend the more critically the TSC will consider the application.
 Budget for this program is a finite resource.
 * Equity: Preference given to individuals who do not have a corporate travel fund and have
--- a/Member-Travel-Fund.md
+++ b/Member-Travel-Fund.md
@@ -1,0 +1,40 @@
+## Purpose
+
+To establish and administer a fund for members of the Node.js
+Foundation to travel and spread knowledge about the foundation
+and the Node.js Foundation projects.
+
+## Restrictions
+
+* Candidates must be Individual Members of the Node.js Foundation.
+* Candidates must be presenting one of the following presentations:
+ * [The Node.js Foundation](https://github.com/nodejs/foundation-slides).
+
+## Process
+
+New slide decks are added to the list of approved decks on an ongoing basis.
+
+* Slide deck must be managed by the Node.js Evangelism WG.
+* Send a pull request to this file adding the slides to
+ the list above.
+* Slide deck must be approved by the TSC.
+
+Any member can apply to give an approved talk. Request may take up to a month
+for the TSC to approve or decline.
+
+* Log an issue in this repository.
+ * Include the name of the event you plan to attend.
+ * Include the location of the event and where you will be traveling from.
+ * Include the presentation you intend to give.
+ * Include the size of the stipend you wish to receive.
+  * Stipends are anywhere from $100 to $1,500 USD and vary with travel distance.
+
+The TSC will consider each proposal in its next regular meeting. While it is
+ultimately at the TSC's discretion who to approve the following considerations
+are made.
+
+* Impact: Preference given to underserved communities.
+* Cost: The larger the stipend the more critically the TSC will consider the application.
+Budget for this program is a finite resource.
+* Equity: Preference given to individuals who do not have a corporate travel fund and have
+not previously received a stipend.


mikeal:
* @nodejs/evangelism because they will be managing the slides.
* @nodejs/tsc for approval.

mikeal:
Note that once the TSC ratifies this it will still need to go to the Board of Directors for approval.

Member-Travel-Fund.md
```diff
@@ -0,0 +1,40 @@
...
+
+To establish and administer a fund for members of the Node.js
+Foundation to travel and spread knowledge about the foundation
+and the Node.js Foundation projects.
+
+## Restrictions
+
+* Candidates must be Individual Members of the Node.js Foundation.
+* Candidates must be presenting one of the following presentations:
+ * [The Node.js Foundation](https://github.com/nodejs/foundation-slides).
```
indutny: Indent.

indutny: Ah, nevermind.

indutny:
LGTM

Member-Travel-Fund.md
```diff
@@ -0,0 +1,40 @@
+## Purpose
+
+To establish and administer a fund for members of the Node.js
+Foundation to travel and spread knowledge about the foundation
+and the Node.js Foundation projects.
+
+## Restrictions
+
+* Candidates must be Individual Members of the Node.js Foundation.
+* Candidates must be presenting one of the following presentations:
```
rvagg: I'd like this wording loosened up a little to make it "based on" or "approved by the TSC"

bnb:
+1.

bnb:
As a member of the Evangelism WG, this seems fine to me. +1, if it counts.

Member-Travel-Fund.md
```diff
@@ -0,0 +1,40 @@
...
+ * Include the presentation you intend to give.
+ * Include the size of the stipend you wish to receive.
+  * Stipends are anywhere from $100 to $1,500 USD and vary with travel distance.
+
+The TSC will consider each proposal in its next regular meeting. While it is
+ultimately at the TSC's discretion who to approve the following considerations
+are made.
+
+* Impact: Preference given to underserved communities.
+* Cost: The larger the stipend the more critically the TSC will consider the application.
```
rvagg: We should also know where the money is going, does it need to be paid to the conference to reimburse or to the individual doing the purchasing. A note that we'll need proof-of-purchase or something official to match the $$ against what was spent would be helpful here too.

Member-Travel-Fund.md
mikeal: Hrm... doesn't this open us up to a lot more administrative overhead in approving every deck that comes through?

rvagg:
Perhaps a note in here targetted at conference organisers too that they can contact us to request a speaker if they'd like one—this is aimed at promotion of the Node.js project and the Node.js Foundation, it's not about the individuals _per se_, it should also be seen as a way for conference organisers to tap into additional resources for mutual benefit.

mikeal:
@rvagg great idea on the 'conferences requesting speakers.' Do we want to make this a slightly different process by which they say they need someone and we allow multiple people to apply to give the talk and then select one?

bnb:
@mikeal The problem with that is certain speakers will be preferred over others by conferences, and get more events - if the selection process were on our side, this would work.

mikeal:
@bnb right, we'd still want the TSC to select the speaker in this case, especially since we'd give preference to a local speaker over an international one.

Member-Travel-Fund.md
```diff
@@ -0,0 +1,40 @@
...
+## Process
+
+New slide decks are added to the list of approved decks on an ongoing basis.
+
+* Slide deck must be managed by the Node.js Evangelism WG.
+* Send a pull request to this file adding the slides to
+ the list above.
+* Slide deck must be approved by the TSC.
+
+Any member can apply to give an approved talk. Request may take up to a month
```
bnoordhuis: `A request` or `Requests`?

Member-Travel-Fund.md
```diff
@@ -0,0 +1,40 @@
...
+ * Include the location of the event and where you will be traveling from.
+ * Include the presentation you intend to give.
+ * Include the size of the stipend you wish to receive.
+  * Stipends are anywhere from $100 to $1,500 USD and vary with travel distance.
+
+The TSC will consider each proposal in its next regular meeting. While it is
+ultimately at the TSC's discretion who to approve the following considerations
+are made.
+
+* Impact: Preference given to underserved communities.
```
bnoordhuis: Very vague.

Member-Travel-Fund.md
```diff
@@ -0,0 +1,40 @@
+## Purpose
+
+To establish and administer a fund for members of the Node.js
```
Fishrock123: What are "members"? Does this mean nodejs org members, or foundation members?

Member-Travel-Fund.md
indexzero: I agree with @rvagg here. Not letting presenters personal flair through on these decks would be a loss imho. I'm happy to volunteer with deck review to help streamline.

Member-Travel-Fund.md
indexzero: +1. Should be:

> Requests may take up to a month

or

> Each request may take up to a month

Member-Travel-Fund.md
mikeal: Individual foundation members https://github.com/nodejs/members/issues/1

Member-Travel-Fund.md
indexzero: Agreed. Helping to support diversity and increasing outreach in underserved communities is very important. As such we should try to be more specific in what areas those are. e.g. the work that @julianduque, and the rest of the JSConf Colombia team is a great example. 

I'm not exactly sure, however, how to change the wording here to express this intent. @mikeal any thoughts? Also summoning @hackygolucky because she is very good at words in this area.

mikeal: The thinking here is that within a subject we're trying to promote awareness about (right now that's just general knowledge about the foundation and the project but we expect that to change over time) there are people, communities and geographies where that knowledge is not very wide spread. For instance, a meeting in San Francisco would be less attractive than a meetup of the same size in Cincinnati. This is especially true when we start looking at non-english speaking countries and non-JS conferences (FOSDEM, Grace Hopper, etc).

This is a bit vague because our understanding of under-served areas is going to evolve over time, as is the scope of the information we intend to persist will grow as we add more slide decks.

Member-Travel-Fund.md
hackygolucky: If proof of purchase is required then are y'all assuming that it will only be -after- the speaker has attended that this stipend would be rewarded? This is a barrier for some--they can't afford to commit to speaking if they don't know whether it will be covered. I see this quite often in underrepresented groups speakers as they are also more likely to be paid less for their day work. I've experienced many a speaker having a tough time even waiting until the conference date in order to be reimbursed. I understand that can make things much more complicated accounting-wise, however.

Member-Travel-Fund.md
indexzero: @mikeal you've got some good language in you comment, how about this? 

> * Impact: Preference given to underserved communities. More specifically, within a subject the foundation trying to promote awareness about (e.g. general knowledge about the foundation) there are people, communities and geographies where that knowledge is not very wide spread. This is where preference will be given. For instance, an event in San Francisco would be less attractive than an event of the same size in Kansas City.

Member-Travel-Fund.md
bnb: @hackygolucky This was the one and only reason I didn't go to the Node summit. I didn't have the means to put up resources (money) as collateral, that would be paid back at a later date. I know I'm an edge case, but I feel like I should voice my angle to show that it does affect people.

mikeal: I'll need to make sure that the foundation can handle a pure stipend (rather than a reimbursement which is what we've done previously). It's actually much easier to manage costs using stipends, people consistently wait until the last minute when the costs go up when they know they are being re-imbursed.

As far as "proof of purchase," we should substitute with something like "link to ticket / hotel prices."

indexzero: @rvagg iirc the plan was that it wouldn't be reimbursements but would piggy-back off of the Linux Foundations travel booking system. @mikeal is that correct? 

Either way we should be explicit about how folks should expect travel to be booked: hotels, flights, etc. Are they expected to lay out the $$$ and get reimbursed? Or is the foundation going to pay for it up-front? This has been important when booking speaker travel in the past for the conferences I've run.

mikeal: > piggy-back off of the Linux Foundations travel booking system. @mikeal is that correct? 

This wasn't specified yet but will need to be specified in the terms of the program before being approved by the board. I've actually never used the LF travel booking system. For ease of managing/budgetting by the TSC I recommend a straight stipend provided the LF can easily send them out (I've only sent them reimbursements thus far).

kosamari:
I assume this is "official" doc for TSC/Foundation.
But once program is in place, maybe evangelism group can make wording friendlier for website ?
something like Hoodie is very encouraging in my opinion.  http://hood.ie/contribute/#talks

mikeal:
@kosamari good point! we should add something about this to the "individual membership" page once it goes up. this is basically a process document so it isn't as accessible as we'd want the documentation to be.

Member-Travel-Fund.md
hackygolucky: +1 this language sounds swell! Is there a reason we're talking underserved instead of underrepresented or is that a separate discussion to be made?

mikeal: > underserved instead of underrepresented

They can be separate but I'd say in this case underrepresented is either a subset of under-served or a by-product of being under-served. What we're trying to address is an information disparity and intervene by spreading that knowledge where we have the opportunity to do so. That disparity exists in several dimensions, some being geographic or language barriers and other being race or gender. Without this information being more accessible to these communities they are less represented because a prerequisite to being represented would be to contribute and these talks do a lot to show where and how to contribute.

Also, by "underserved" we mean not only that an information gap exists but that it is not "served" by the existing speakers already available to these events without our intervention.

In terms of this program, we can only address what we are presented with the opportunity to address (people have to apply to the program, and there are only so many opportunities to speak especially in areas we know are under-served).

However, we make sure the program isn't too specific so that it can address new issues as they arise. If, for instance, we want to create a deck about libuv targeted at getting more contributors involved in that project we may want to adjust the communities we target based on other criteria (existing Windows developers, for example).

Member-Travel-Fund.md
mikeal: We should discuss this a little more.

I'm basing this a bit on the success we had with the io.js Roadmap, which was translated in to like 11 languages and given at I don't even know how many meetups. One of the things we got out of that was collaborative iteration of the deck and an "approved" resource that encouraged localization.

I gave this Foundation talk at FresNode last week and there was plenty of "personal flair" in the way I delivered it even if the deck lacked personality. However, I do like talks that are more personal, and add a bit about the persons involvement, and I certainly touched on that during my delivery but didn't have anything extra in the deck to support it visually.

PR #25, opened 2015-12-20
Add Proposal-Process.md

Trott:
Add document that describes how the foundation handles proposals to
create or change a process foundation-wide.

This came out of https://github.com/nodejs/inclusivity/pull/68. The idea there is to document how this stuff works because the proposed Inclusivity WG is not expected to have any power whatsoever to enact changes to process to any other WGs. The question that needed answering is: If we want to propose a change that would apply foundation-wide, how do we propose the change and what is the process for the TSC (and/or other governing body of the foundation) to approve or reject the change?

I offered to put together a minimal doc that at least would give people something to talk about along the lines of "The twelve things missing from this document are...." This is that document.

@mikeal commented:

> Not to get too meta, but this is probably something we want to put on the TSC repo and expand on at some point 

I figure there's no time like the present, so here it is. It could probably use a decent title, some mention of a timeline for the process, and some indication of the elements that need to be included in a proposal. But those can be fleshed out over time. Hopefully what's here is mostly uncontroversial (as it's basically me paraphrasing @mikeal and @jasnell) but if I did get anything wrong, I'd rather know now than having it become part of the Inclusivity repo without getting corrected.


Proposal-Process.md
```diff
@@ -0,0 +1,14 @@
+For a proposal to create or change a process foundation-wide, a PR should be
```
jasnell: "foundation-wide" is misleading. The TSC does not establish policies for the Foundation, which is a legal entity with it's own governing board of directors. The TSC establishes policies for the technical governance of the Top Level Projects that exist under the Foundation, yes, but that's not the same as "foundation wide".

jasnell: Minor nit: Can you add a header to this document? `## Policy Change Proposal Process` or something along those lines

jasnell:
Overall this is fine, but why not simply add this to the README.md?

Proposal-Process.md
Trott: `foundation-wide`: I originally struggled with what to put there and came up with `organization-wide` and `project-wide`. Neither of those felt *quite* right but I went with them on the theory of "just get *something* out there and let other people comment on it already".

Without noting my own misgivings about the terms, @mikeal flagged them as not accurate and (implicitly, perhaps) suggested `foundation-wide`. Here's Mikeal's' comment:

> The problem with "project" and "organization" wide is that it could be interpreted as "just Node.js" or "just the Node.js GitHub org" when what I believe we want to do is implement these "foundation-wide" so that they apply to any top level project we bring in.

(If you want more context: https://github.com/nodejs/inclusivity/pull/68#discussion_r48095093)

I'm happy to change the wording to something else, but I would rather be told what that wording is. I'm struggling coming up with good wording.

TL;DR: We've tried each of the following and none have proven satisfactory. What is the wording we should use instead?

* foundation-wide
* project-wide
* organization-wide

Trott: Adding a title: That's the other thing I struggled with...what to call this thing. Does `Policy Change Proposal Process` imply that this only applies to "policies" (whatever those are) and not "guidelines" or "charters" or whatever?

I'm probably over-thinking this and will go with your suggestion, @jasnell.

Trott:
Adding to the README sounds like a good idea to me. I'll go ahead and do that.

Proposal-Process.md
jasnell: Perhaps something like:

```
The Node.js TSC is chartered to oversee the technical governance of all Top Level Projects under the Node.js Foundation. Accordingly, the TSC establishes the default governance, conduct, and licensing policies for all Top Level Projects. To propose changes to those policies, a PR should be opened in the `nodejs/TSC` repository...
```

Trott: Works for me. I'm inclined to specify Working Groups too because (if I understand correctly) not all Working Groups need to fall under a particular Top Level Project.

Trott:
I added some more explanatory text, some of it from @jasnell and some if it of my own doing.

mikeal:
LGTM.

README.md
```diff
@@ -21,3 +21,28 @@ If your project is interested in joining the Node.js Foundation please read the
...
+
+## Policy Change Proposal Process
+
+The Node.js TSC is chartered to oversee the technical governance of all Top
+Level Projects and Working Groups under the Node.js Foundation. The TSC
+establishes the default governance, conduct, and licensing policies for all Top
+Level Projects. Top Level Projects and Working Groups have broad powers of
+self-governance.
+
+To propose a change or addition to policies or prcesses that are intended to
```
jasnell: s/prcesses/processes

README.md
```diff
@@ -21,3 +21,28 @@ If your project is interested in joining the Node.js Foundation please read the
...
+cover all Top Level Projects and Working Groups in the foundation, a PR should
+be opened in the `nodejs/TSC` repository. 
+
+The pull request can be labeled `tsc-agenda` to request that it be put on the
+agenda for the next TSC meeting.
+
+The Node.js Foundation Board of Directors retains certain rights (especially
+legal considerations). If the TSC endorses a proposal, they will escalate to the
```
jasnell: Perhaps, `If the TSC endorses a change proposal for any policy or guideline that falls under the responsibility of the Node.js Foundation Board, the TSC will escalate...`

jasnell:
LGTM with a couple additional nits

README.md
mikeal: "falls under the responsibility" is too narrow. The TSC also escalates when its asking for advice or direction from the board. Our recent questions about how to store copyright information is a good example, its on the line between technical and legal so we're asking for advice even though it's the TSC's responsibility to maintain that copyright information.

Trott:
Typo fixed. I think that's the only outstanding nit...

jasnell:
LGTM

jasnell:
Landed in 8349899

PR #24, opened 2015-12-19
doc: add Working Groups document

ashleygwilliams:
this document adds a `WORKING_GROUPS.md` file based entirely on the file that lives in [nodejs/node](https://github.com/nodejs/node/blob/master/WORKING_GROUPS.md) as per [@mikeal's suggestion](https://github.com/nodejs/node/pull/4355#issuecomment-166020339) on the [PR to add the nodejs/inclusivity working group](https://github.com/nodejs/node/pull/4355)

the only change to this doc from the [nodejs/node one](https://github.com/nodejs/node/blob/master/WORKING_GROUPS.md) is that i referenced it in this one.

PR #26, opened 2015-12-20
doc: add WG "bootstrap kit"

jasnell:
Separate out the templates for bootstrapping a WG process...

Related to #24 . This is not intended as an alternative to #24 but an adjunct to it. The idea is to separate things out a bit more to make the process easier to maintain and update. We ought to be able to point the nodejs/node WORKING_GROUPS.md file to these templates also as opposed to maintaining to separate set of bootstrap templates in the two separate WORKING_GROUPS.md files.

/cc @mikeal @sup @ashleygwilliams @nodejs/tsc @nodejs/ctc


Trott:
Yes! Issues will shake out as people use it, but it only needs to be good enough to be usable and not perfect or complete. Yes!

sup:
other than what i outlined in my comments (which are nitpicks, sort of), i think it's a good thing! +1 from me

Trott:
All occurrences of the word `weekly` appear to be in template text, so I think the idea is that WGs can edit it as appropriate for their group. It's just offered as a starting point.

sup:
@Trott ah, makes sense

BootstrapKit/CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,14 @@
+## Working Group Code of Conduct Template
+
+Every Working Group must have a Code of Conduct. By default,
+all repositories under the Node.js GitHub Organization fall under
+the [Node.js Code of Conduct][], but Working Groups can choose to
+define their own alternative policy. If the Working Group does choose
+it's own Code of Conduct, it must be documented in a `CODE_OF_CONDUCT.md`
+file located in the root of any repository under their stewardship.
```
Fishrock123: The CoC can't do less than the Node.js Code of Conduct. Or shouldn't be able to.

i.e This can't be used to bypass one, or if we overlooked that, we need to change things so it can't be.

BootstrapKit/Charter.md
```diff
@@ -0,0 +1,28 @@
...
+Every Working Group requires a charter that describes the scope of
+the WGs responsibilities. Once written, a new Pull Request that adds
+the charter to either the [CTC][] or [TSC][] WORKING_GROUPS.md document
+must be opened. The WG is considered to be officially chartered once the
+PR adding the proposed charter is landed and merged into the relevant
+WORKING_GROUPS.md document
+
+### Template: *[insert WG name]* Working Group
+
+The Node.js *[insert WG name]* is jointly governed by a Working Group (WG)
```
nebrius: Grammar nit/question: the phrase "jointly governed" implies that two or more entities govern this WG, but the phrase only lists one entity: the WG. Should we add something like "is jointly goverened by a WG...and the TSC" or something to that effect?

Fishrock123: "WG Members"?

BootstrapKit/Governance.md
```diff
@@ -0,0 +1,140 @@
...
+
+WG seats are not time-limited.  There is no fixed size of the WG.
+However, the expected target is between 6 and 12, to ensure adequate
+coverage of important areas of expertise, balanced with the ability to
+make decisions efficiently.
+
+There is no specific set of requirements or qualifications for WG
+membership beyond these rules.
+
+The WG may add additional members to the WG by unanimous consensus.
```
nebrius: Question on the word "may", here. Is the intent that WG's have the option of requiring unanimous consensus for membership, but it's not required? Or is this saying that adding new members can only happen by unanimous consensus, and "may" means that WG's have the option of adding members?

If it's the latter, perhaps we could change the end to "WG by unanimous consensus only"?

BootstrapKit/Governance.md
```diff
@@ -0,0 +1,140 @@
...
+membership decisions.  If a WG member is unable to attend a meeting
+where a planned membership decision is being made, then their consent
+is assumed.
+
+No more than 1/3 of the WG members may be affiliated with the same
+employer.  If removal or resignation of a WG member, or a change of
+employment by a WG member, creates a situation where more than 1/3 of
+the WG membership shares an employer, then the situation must be
+immediately remedied by the resignation or removal of one or more WG
+members affiliated with the over-represented employer(s).
```
nebrius: Would it be advantageous to specify a maximum length of time that this can happen?

I would imagine removal won't happen right away, especially if there's disagreement over who should leave. It could also be the case that there is someone interested in joining, and fast-tracking that membership would get the one employer under the 1/3 mark again. Either approach takes time, and I think it would be unfair to require this to happen imemdiately, but we also want to make sure that WG's don't use the "we need more time" as an excuse for inaction and not fixing the issue

BootstrapKit/Governance.md
```diff
@@ -0,0 +1,140 @@
...
+
+Any community member or contributor can ask that something be added to
+the next meeting's agenda by logging a GitHub Issue. Any Collaborator,
+WG member or the moderator can add the item to the agenda by adding
+the ***WG-agenda*** tag to the issue.
+
+Prior to each WG meeting the moderator will share the Agenda with
+members of the WG. WG members can add any items they like to the
+agenda at the beginning of each meeting. The moderator and the WG
+cannot veto or remove items.
```
nebrius: Sometimes, I've seen it be necessary to remove items during meetings themselves for time reasons (when other discussions go over). What do you think about changing the last sentence to something like "Agenda items can only be removed by consensus of all members in attendance"?

BootstrapKit/ModerationPolicy.md
```diff
@@ -0,0 +1,14 @@
+## Working Group Moderation Policy Template
+
+Every Working Group must have a Moderation Policy. By default,
```
nebrius: I liked the phrasing you used for the CoC:

> Every Working Group must have a Code of Conduct. By default, all repositories under the Node.js GitHub Organization fall under the [Node.js Code of Conduct][]

Would it make sense to reuse that wording here? Make it known that one is required, even if it's just a link to the Node.js one?

mikeal: Yes, this is ideal. Keep in mind tho that this language is a departure from our prior policies (which are kind of a failure in this regard so +1 to the change).

BootstrapKit/Governance.md
mikeal: So, the Website WG operates without meetings. This is actually going quite well. If at all possible I think WGs *should* not have meetings. The only reason to have a meeting is if you can't be effective without one, and there are plenty of WGs that fall under that, but I'm worried about having meetings the base policies because I've seen this anti-pattern of "ok, we have a WG now, let's have a meeting" and then all the momentum is lost getting the meeting going.

BootstrapKit/Governance.md
mikeal: I think this is the right approach for the base policy. *Some* WGs will want to change this and iterate on it. I think for certain groups, like LTS and CTC, we might want to have rules that ensure every members is actively engaged in the code base and if they take a break from the code base they, temporarily, step off the governance body. As long as we make it simple and fluid to come on and off that approach might be best for certain groups. However, I wouldn't want that to be in the base policy. Also, keep in the mind that this currently mirrors the language and terms in the TSC charter, which might be good for continuity, and we can't change the TSC Charter without board approval.

BootstrapKit/Governance.md
mikeal: This is ancient language that should be removed. Nothing should have unanimous consensus, it's ineffective and potentially poisonous :)

BootstrapKit/Charter.md
mikeal: Ya, I think the term "jointly" refers to the membership of the WG *being* the WG governance body.

BootstrapKit/CODE_OF_CONDUCT.md
mikeal: How do you define "bypass" though? That's really hard and you're opening up the language in every lower level CoC to be under a strange bikeshed where people re-interpret the language in the base CoC.

There's really only two ways you can phrase this to be enforcable and sustainable:
* Make the base CoC apply to all WGs and TLPs not matter what, allowing only *additional* terms to be added by independent bodies.
* Allow the CoC to be replaced entirely and if the TSC is upset enough about it they can yank the charter (extreme and doubtful but is quite literally the fallback mechanism we have for all other policies and it seems to be working so far).

mikeal:
My apologies in advance, the current *actual* state of governance is a little behind what is documented. Additionally, we've been talking about a variety of improvements because some of these approaches haven't worked. Unfortunately this means that when you do the good work of documenting things properly you're going to get far more feedback than is fair :/

First off, I don't think we should use the term "bootstrap." We need to move away from that as it has been something of a failure. Most new working groups do not copy/paste the necessary policies. Also, those policies end up in a state of disrepair as improvements are made to the original policies and never make their way into the copies.

I would call the directory simply "Policies." In the README and in the policies we should move a "copy on write" process. Basically, when you spin up a new WG you should just reference the base policies in your README and copy the CONTRIBUTING.md file (this is an unfortunate requirement because the DCO is in there and that file is *special* and referenced when sending a PR).

If a WG or TLP wants to iterate on a policy they should copy it and edit it with PRs, running those changes through their standard governance process. If they haven't copied it they fall under the base policies. This is a far more sustainable practice. It also makes it infinitely easier for us to iterate on base policies as we learn.

mikeal:
In case it wasn't clear in my inline comment, I'm +1 on removing the entire section about meetings.

BootstrapKit/Charter.md
nebrius: Ah I see, that makes sense. Perhaps we could go with something like "The Node.js *[insert WG name]* is jointly governed by the members of the WG" to make that more clear?

BootstrapKit/Governance.md
nebrius: kk, that makes sense

BootstrapKit/Governance.md
jasnell: Well, how about removing members?

jasnell:
Ok, pushed a commit that makes several changes:

1. It renames the directory to `BasePolicies`
2. It notes that the WG Meetings section of the Governance document is optional
3. It adds language to the CODE_OF_CONDUCT.md and ModerationPolicy.md files to indicate that the policies should not be weaker or too divergent from the default policies

BasePolicies/Governance.md
```diff
@@ -0,0 +1,149 @@
+## Working Group Governance Template
+
+Every Working Group must have a documented governance model that describes the
+basic operation of the Working Group. This must be documented in the repository
+and referenced by the `README.md`.
+
+## Template: *[insert WG name]* Working Group Governance
```
nebrius: Just for clarification, this is CONTRIBUTING.md, correct? Would it make sense to explicitly mention this template goes in CONTRIBUTING.md?

jasnell: Perhaps but I'm not sure that's a hard requirement.

jasnell:
@nodejs/tsc ... can I ask for another review on this?

BasePolicies/Governance.md
```diff
@@ -0,0 +1,149 @@
...
+WG on an ongoing basis.
+
+Individuals making significant and valuable contributions are made
+Collaborators and given commit-access to the project. These
+individuals are identified by the WG and their addition as
+Collaborators is discussed during the weekly WG meeting.
+
+_Note:_ If you make a significant contribution and are not considered
+for commit-access log an issue or contact a WG member directly and it
+will be brought up in the next WG meeting.
```
Trott: This paragraph seems like it could use a bit more polish. Maybe something like this?

> Individuals who have made significant contributions and who wish to be considered for commit-access may create an issue or contact a WG member directly. It is not necessary to wait for a WG member to nominate the individual.

BasePolicies/Governance.md
```diff
@@ -0,0 +1,149 @@
...
+responsibility for the change. In the case of pull requests proposed
+by an existing Collaborator, an additional Collaborator is required
+for sign-off. Consensus should be sought if additional Collaborators
+participate and there is disagreement around a particular
+modification. See _Consensus Seeking Process_ below for further detail
+on the consensus model used for governance.
+
+Collaborators may opt to elevate significant or controversial
+modifications, or modifications that have not found consensus to the
+WG for discussion by assigning the ***WG-agenda*** tag to a pull
```
Trott: The first sentence in this paragraph is long and confusing. Minimally, it should get some improved clarity via punctuation. I'd probably just decide that "controversial" and "have not found consensus" are mostly the same thing and just use one or the other:

> Collaborators may opt to escalate significant or controversial modifications to the WG as a whole for discussion by assigning the...

BasePolicies/README.md
```diff
@@ -0,0 +1,42 @@
+## Establishing a new Working Group
+
+A Working Group is established by first defining a charter that can be
+ratified by the TC. A charter is a *statement of purpose*, a
+*list of responsibilities* and a *list of initial membership*.
```
Trott: Nit: The Charter template does not seem to include a statement of purpose or a list of responsibilities. It would be ideal if the charter template lined up very clearly with the elements of a charter described here. (But I wouldn't hold this up just for that. It can be done as a subsequent improvement.)

Trott:
Nit: Some files are in ALL_CAPS_WITH_UNDERSCORES and some are NotInAllCapsWihtUnderscores. Might be nice to be consistent, allowing for `README.md` to be an exception (along with a few others, like `CONTRIBUTING.md` which actually might be nice to have some boilerplate text for).

jasnell:
@Trott .. thank you. I'm about to push a few edits based on the feedback. The difference in the uppercase vs. lowercase file names is to maintain consistency with existing documents, which, unfortunately have been handled inconsistently up to this point :-)

jasnell:
@Trott @sup @nebrius @juliepagano @mikeal @node/tsc ... I'd like to go ahead and get this landed. I'm sure there's more edits that can be made but can I get a couple LGTM's to get it landed?

sup:
LGTM!

nebrius:
LGTM too!

jasnell:
Landed in b0533ea8

ashleygwilliams:
sorry i didn't have the bandwidth to participate on this, but looking at it now and i think it is great :) thanks for this @jasnell

PR #29, opened 2016-01-08
add inclusivity working group

ashleygwilliams:
fixes: https://github.com/nodejs/inclusivity/issues/69

### DIFF
--- a/WORKING_GROUPS.md
+++ b/WORKING_GROUPS.md
@@ -17,16 +17,39 @@ should be dissolved and the responsibility for governance absorbed back into
 the TSC. A Working Group can be dissolved either through consensus of the
 Working Group membership or normal TSC motion and vote.
 
-## Current Working Groups
-
-> (none)
-
-Technical Working Groups exist in the [nodejs/node repository](https://github.com/nodejs/node/blob/master/WORKING_GROUPS.md).
-
-#### Process:
+## Process:
 
 * [Starting a Working Group](#starting-a-wg)
 * [Bootstrap Governance](#bootstrap-governance)
+
+## Current Working Groups
+
+NOTE: Technical Working Groups exist in the [nodejs/node repository](https://github.com/nodejs/node/blob/master/WORKING_GROUPS.md).
+
+* [Inclusivity](#inclusivity)
+
+### [Inclusivity](https://github.com/nodejs/inclusivity)
+
+The Inclusivity Working Group seeks to increase inclusivity and diversity
+for the Node.js project:
+  * *Increasing inclusivity* means making the Node.js project a safe and
+    friendly place for people from diverse backgrounds.
+  * *Increasing diversity* means actively onboarding people from diverse
+    backgrounds to the Node.js project and maintaining their participation.
+
+Its responsibilites are:
+
+* Foster a welcoming environment that ensures participants are valued and can
+feel confident contributing or joining discussions, regardless of any aspect of 
+their identity.
+* Proactively seek and propose concrete steps the project can take to increase
+inclusivity.
+* Serve as a resource for the development and enforcement of workflows that 
+protect community members and projects from harassment and abuse.
+* Acknowledge and celebrate existing diversity accomplishments within the
+project while seeking to build upon them.
+* Identify ways to measure diversity and inclusivity within the project and
+report them at regular intervals.
 
 ## Starting a WG
 


jasnell:
LGTM

jasnell:
@nodejs/tsc

mikeal:
Looks great :)

rvagg:
A couple of different bullet-point styles in use there (`-` and `+`, indented and non-indented first line, aligned and non-aligned subsequent lines), would you mind picking one pls @ashleygwilliams cause this will set the format for future additions to the doc.

I see we have an unfortunate mix of >2 styles in the core repo WORKING_GROUPS.md, perhaps we can pick a style here and update there afterwards.

ashleygwilliams:
totally @rvagg! good feedback. i was actually copying what i saw in the core `WORKING_GROUPS` hahaha, hence the multiple styles!

any preference? (i have none except preferring consistency, which it seems we share :smile:)

jasnell:
Lol yeah, we gotta fix the styles in that one too

rvagg:
Looking over other top-level .md files I think I'm seeing this format showing up the most, although there is a bit of variation:

```
* Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor
  incididunt ut labore et dolore magna aliqua.
* Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip
  ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit
  esse cillum dolore eu fugiat nulla pariatur.
* Excepteur sint occaecat cupidatat non proident.
```

This is the style I personally use but in this case I only care about consistency.

@nodejs/documentation suggestions for a style to adopt? Perhaps we have a prevailing style in the docs?

ashleygwilliams:
wfm. i'll update + amend.

ashleygwilliams:
oh whoops, missed your ask of @nodejs/documentation. will hold off til tomorrow.

rvagg:
@ashleygwilliams my vote for now would be to just adopt a consistent style and we'll do a general cleanup later across files. I'd love if @nodejs/documentation could head that but I'm not sure they are organised enough yet.

ashleygwilliams:
fair enough! incoming.

ashleygwilliams:
updated this @rvagg ICYMI. if yer just busy or working on something else, no worries!

jasnell:
Still LGTM :-) ... one minor nit tho @ashleygwilliams ... the `#### Process:` header depth ends up causing it to be nested under the Inclusivity WG charter. Can you either move it up a level by dropping one of the `#`'s or move the `#### Process` section up above the Inclusivity Charter itself?

ashleygwilliams:
ah, excellent catch @jasnell -- do you have a preference re: solution? i think removing the `#` is best, personally, but i don't have strong feelings.

jasnell:
I'm good with either. Begone with the #! ;)

ashleygwilliams:
incoming.

jasnell:
Excellent. Thanks much. I'll wait for @rvagg to weigh in before landing... but will go ahead and land tomorrow if there are no objections posted.

ashleygwilliams:
heh, actually, i moved it up and made it a `##`, looks like this now, which i think is much clearer than before. thoughts, everyone? cc/@jasnell
screen shot 2016-01-13 at 4 38 35 pm

jasnell:
That works too!

rvagg:
added `tsc-agenda` label, I'm not sure if it's been through a meeting yet (I missed at least the last one), should be cleared up by tomorrow if we can get a quorum

ashleygwilliams:
thanks for the update @rvagg !

jasnell:
This was just discussed on the TSC call. There were no objections! We should be good to go. I'll get this landed shortly.

mikeal:
Ratified in the TSC meeting on Jan 14th.

sup:
WOOT

nebrius:
🎉

zkat:
🎉👏🍆!!

jasnell:
@zkat .. hmm.. just a side note: given the issues we've had with having to moderate users who post comments with eggplants, it's likely not the best choice to include one in https://github.com/nodejs/TSC/pull/29#issuecomment-171777577. I know your intent is not to troll, but it would be rather hypocritical of us to not apply the same moderation criteria to your post as we do others :-)

PR #34, opened 2016-01-21
add initial TLP application for the Appium project

jlipps:
[Appium](http://github.com/appium/appium) wishes to join the Node Foundation as a top level project. Here is our application.

cc: @mikeal @imurchie @bootstraponline


jasnell:
@jlipps ... thank you for this. Will be taking a look to review shortly. Going to proactively tag this with `tsc-agenda` however to get it on the next TSC meeting agenda

jlipps:
@jasnell excellent, thanks. Will wait to hear more!

Applications/Appium-TLP.md
```diff
@@ -0,0 +1,182 @@
...
+* Milestone tracking: [GitHub](https://github.com/appium/appium/milestones)
+* Releases: [GitHub](https://github.com/appium/appium/releases) and [NPM](https://npmjs.org/package/appium)
+* User discussion forums: [discuss.appium.io](https://discuss.appium.io)
+* User chat: [gitter.im/appium](https://gitter.im/orgs/appium/rooms) (#appium on Freenode is mostly abandoned)
+* Developer/contributor chat: [appium.slack.com](https://appium.slack.com)
+* Developer mailing list: [appium-developers@googlegroups.com](https://groups.google.com/forum/#!forum/appium-developers)
+
+### Intellectual Property and Other Assets
+
+Appium does not currently have any registered trademarks. We own/manage the
```
mikeal: By "we" do you mean the community or a company? We need to know who to talk to about transferring the assets if you get admitted :)

jlipps: Good question. In terms of the domain, Sauce currently has it. In terms of the logo, I'd say it's technically the same given that it was designed by a Sauce employee. So we should maybe officially release that as well in a transfer.

jlipps: What makes sense then in terms of how to update these sentences for the purpose of this application?

mikeal: Just note the assets and who currently owns them so that at some point in the future the TSC could hand that off to the foundation lawyers to go and have transferred.

jlipps:
@mikeal updated the IP section

rvagg:
@jlipps you'll have to be patient with us while we work through our goals and general philosophy regarding top-level projects, I'm sure you'll understand that this is all new for us and we're still trying to find our footing before being able to make clear decisions. We'll keep you updated as we make progress though.

bnoordhuis:
I haven't commented before but seeing how it's going to be discussed at today's TSC meeting: it's not quite clear - at least, to me - what you expect from the Node.js Foundation.

I also don't quite see how Appium fits in with other top-level projects; 'incongruous' is the word that comes to mind.

jasnell:
One quick point: it's possible to let this incubate for a while without making a decision yet about if this should become a top level level. The incubation phase is specifically intended to allow everyone time to figure things out. I'm +1 on incubating Appium but haven't yet made a decision about whether I think it makes sense as an official TLP

jlipps:
@bnoordhuis @jasnell fair question. Our main motivation is to transition Appium from corporate stewardship to non-profit ownership and open governance. We like how Node.js has made decisions in these areas and feel we would benefit from insight (and oversight) as regards running a truly open project. Given Sauce Labs's involvement with the Node Foundation (we're partners/sponsors or whatever), and a random discussion I had with @mikeal, it was the first place I wanted to turn.

One way to look at Appium vis a vis other actual or possible TLPs is "incongruous"; I get that. Another way to look at it is a demonstration of the breadth of Node.js as a platform. Appium could have chosen any language but we chose JS/Node. We have avidly kept up with developments in the language and the community. We are pushing Node patterns to depths most open source projects don't encounter because of the size/complexity of our project. We're introducing many new contributors to Node who previously only had experience with Java or Ruby, for example. In short, while we're by no means close to Node core or of likely interest to people who work on Node itself, we would be a symbol of the diversity of what you can do with Node. Does that make it appropriate as a Node TLP? That's up to you all!

If we don't become a Node TLP we will reach out to the Linux Foundation directly or explore other options (I am not an expert on this so will obviously have to do a bunch of research). In that sense we don't _need_ to be a Node TLP. But we think it would be really awesome!

Looking forward to hearing what you all discuss at the meeting. And thanks for setting the expectation that there's a lot of philosophical groundwork to do to figure out whether this is possible.

jasnell:
@jlipps .. thanks for that. As I said, I'm all for getting Appium into the incubation process and really just seeing how things evolve from there. We definitely should take our time and not rush things unnecessarily but I'm happy to help in whatever way we can. 

I think a large part of the hesitation here is that we're all still getting our heads around the precisely why the foundation should grow to include additional top level projects. My thinking on this has evolved and crystalized around simply providing a home for projects that support the node community and have the support of the node community. We may not have all the boxes ticked off but that's ok. Let's just try and see how things evolve. In other words, this has my LGTM

PR #42, opened 2016-02-04
wip In-The-Context-Of-Node.md

Fishrock123:
_Sorry for this still being mostly a very raw spewing of thoughts, this is still mostly WIP since I'm a slow writer, but it'l be brought up at today's meeting anyways so I might as well just publish it anyway._

The goal is to have a document that lays out some of how and why Node is where it is today, to use that for context in deciding what may be important to us, for spending resources on.


mikeal:
One thing that is hard to express without visualizations is the interconnection of many of these parts.

Sometimes is people that connect them, sometimes it's a history of some kind, and sometimes it is quite literally that they all depend on the same thing.

I think stories are a good way to go about that. Like the story of how NodeSchool started and scaled is also a story about why NodeConf and OPEN Open Source have been important.

williamkapke:
@Fishrock123 I think some of this you may have documented elsewhere now? (the Streaming info)

jasnell:
Closing due to lack of forward movement on this. Can reopen if necessary

PR #53, opened 2016-03-01
Add guidance to proposal section

mhdawson:
Add additional guiance to proposal section to call out
that proposals need to explain how the project will
better the Node.js ecosystem.

### DIFF
--- a/Project-Lifecycle.md
+++ b/Project-Lifecycle.md
@@ -106,8 +106,8 @@ case.  A few areas you may want to include:
 * how will the project help the Node.js ecosystem grow as a whole ?
 * what set of Node.js users is the project interesting/critical to ?
 * is there a direct relationship between the project and core Node.js
-  or one of the existing top-level projects?  For example is it a
-  Node.js dependency ?
+  or one of the existing top-level projects?  For example, does Node.js
+  itself depend on this project ?
 * how does the project support or enhance existing top level
   projects ?
 * how would the Node.js ecosytem be affected if the project
@@ -118,6 +118,8 @@ Working Group must include:
 
 * Introduction and project description.
 * An explanation of how the project betters the Node.js ecosystem.
+* Why does the project want to join the foundation and what do you
+  expect to get out of joining.
 * Project history.
 * Any available metrics or even estimates about the user base, ecosystem
 and community.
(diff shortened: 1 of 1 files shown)


feross:
:+1:

feross:
I'd also like to see a bit about why a project would want to join the foundation, i.e. what they can expect from the process and why they might want that.

Project-Lifecycle.md
```diff
@@ -98,10 +98,26 @@ vote to abstain from representation on the TSC.
...
+For a project to be accepted as a top-level project under the Node.js
+Foundation it must help better the Node.js ecosystem.  It is important
+that proposals include a clear explanation of why this will be the
+case.  A few areas you may want to include:
+
+* how will the project help the Node.js ecosystem grow as a whole ?
+* what set of Node.js users is the project interesting/critical to ?
+* is there a direct relationship between the project and core Node.js
+  or one of the existing top-level projects?  For example is it a
+  Node.js dependency ?
```
orangemocha: I wonder if "is it a Node.js dependency" might be misinterpreted. Suggestion "does Node.js itself depend on it"

orangemocha:
Nit: `guiance` => `guidance` in the commit message.

orangemocha:
This looks like a useful starting point to me. LGTM, with a couple of really tiny nits.

mikeal:
These are some good questions.

The state of the docs here need a big re-working to get in-line with what we're already doing. It's very hard to figure out the TSC is doing in practice and what the Incubator is for.

There's also a few tough questions we need to address that I'll be logging over the next few days.

This is all just to say that I think that we should merge this but shortly after re-work the README, this doc, and most of the other docs as well.

mhdawson:
Updated to address review comments so far.

jasnell:
LGTM

mhdawson:
Can this be landed ?

jasnell:
Should be safe but let's put it on the tsc-agenda for tomorrow just to be extra safe ;-) I'll land it tomorrow if there are no objections.

Fishrock123:
TSC has decided to delay this for now while we figure some mission-y things out.

mhdawson:
Don't think this is relevant anymore.

PR #56, opened 2016-03-02
Updating current mentors list.

mikeal:


README.md
```diff
@@ -32,7 +32,11 @@ at their own discretion.
     help projects adopt and iterate on policies and processes that achieve
     these goals and eventually allow them to graduate the incubation phase.
 
-      * Mikeal Rogers (@mikeal)
+      * Mikeal Rogers (@mikeal) Currently assigned to Express.
```
orangemocha: missing @Fishrock123  ?

mikeal: Jeremiah isn't assigned as a mentor to Express, he's an active contributor there.

orangemocha: Ah, I didn't realize that.

README.md
```diff
@@ -32,7 +32,11 @@ at their own discretion.
...
+      * James Snell (@jasnell) Currently assigned to Express.
+      * Rod Vagg (@rvagg) Currently assigned to libuv.
+      * Alexis Campailla (@orangemocha) Currently assigned to libuv.
+
```
orangemocha: I think it would be better to remove the @ character, otherwise people will get notifications should this repo get forked. I would suggest using the same format as in https://github.com/nodejs/node#collaborators

jasnell:
LGTM

README.md
joshmanders: As far as I know @ character does not ping a user in a `.md` file. It's just text there.

orangemocha: I am not sure how this works in detail. I know that I got tons of notifications from PRs like https://github.com/marekventur/dependency-updater/pull/21 because my alias is in the readme. 
In any case, it's not a huge issue either way :)

orangemocha: That might be a different issue though. Those PRs have my alias in the description, not the readme.

joshmanders: Yeah your alias is in the issue body as a comment, which does ping the user and pull them into the issue as a subscriber, but in a repo `README.md` file it won't do that. Hope this clears it up for you. :)

orangemocha: Sure, thanks :) I remembered that we avoided it on the README of nodejs/node, but I don't remember the details of why we did that.
In any case, it's not a big deal and we can fix it later.

And I assume you didn't get a notification for https://github.com/orangemocha/test/commit/9b51d8aa674de6bca3e2d785cd6099fbaaa2b6ce :)

joshmanders: I did not. :)

orangemocha:
LGTM

Issue #54, opened 2016-03-02
Can we effectively define a "project"


mikeal:
This has been troubling me lately.

We've been creating scopes of work as individual "projects." For a few scopes this is simple and effective, but for many it's very inadequate.

Currently, we have 2 groups at the "top level."
* Node.js Core (managed by the CTC)
* Inclusivity Working Group

We then have two other projects that are in the Incubator and *could* come to the top level at some future point in time.
* Express
* libuv

The scope of the Inclusivity WG is pretty understandable. There's certainly work that they will take on in the future that they aren't doing yet but what would be in or out of that group is pretty easily understood. Similarly, libuv has very well defined scope and a single GitHub org.

Where things get tricky is Node.js Core and Express. Here's a quick breakdown

**Node.js Core**
* Single GitHub org w/ ~100 repos.
* 85 teams, about half of which are localization groups.
* Sub-projects which are depended on *outside* of Node.js Core like readable-stream.
* Several "Foundation Level" resources like the website, several micro-sites, help repo, and moderation are actually in the "Core Project."
* `nan` which does not ship with Core and is in-fact depended on solely by external projects.

The connection to "Node.js Core" is not always clear. While `nan` is considered "part of the platform" as it's our preferred native abstraction it is effectively independent.

Several resources in the core project are potentially scoped to the entire ecosystem and not just core. The Documentation WG is working on resources far beyond core and the website has a global event map, etc.

Express is similarly large in scope. It's 3 GitHub organizations worth of repositories, many of which are depended on by other web frameworks as well as Express.

We've been hesitant to change a lot of this structure because, well, it's working. But now that we're considering admitting other projects this complicates matters a bit. For instance, would a framework like Sails or Kraken get its own "project" or would it be put under Express?

As a thought experiment, imagine how these projects would fit into this structure:

* serial-port
* Nodebots
* npm
* request

Now think about the issue of representation to the TSC. If the TSC is made from reps from each project then the TSC ends up being a pretty poor representation of what is going on in the foundation because of the huge disparities in size and scope. It also means that there's no way to break some scope off a top level project into another top level project without it growing the size of the TSC -- which I think we all agree would become problematic.

From the beginning we've worked to guarantee autonomy for individual groups. This has helped to raise incentives for participation by letting contributors own their processes and governance, but another part of its success is that the groups get to have their own identity. If you have to say something like "serial-port would be a sub-group in the NodeBots project" it has an effect on the *identity* of that project.

[...]

Issue #58, opened 2016-03-04
Splitting mentorship from long term project hosting.


mikeal:
One of the things the TSC has struggled with is to try and define the scope of what projects could fit under the TSC.

Simultaneously, the TSC created a mentorship program (the incubator) and has extended those mentorship resources to critical projects in need.

Another aspect of this mentorship has been a neutral home (the Node.js Foundation) which helps when building up a contribution base around the project.

However, the long term implications of hosting *graduated* projects are outside of what I believe the TSC is comfortable maintaining, based on the last meeting.

I'd like to use this issue to discuss any concerns with splitting the long term hosting of graduated projects from the TSC. There are several options for where to host these assets and any long term governance concerns we might have can be structured into where and how they are hosted. It also almost goes without saying that whatever this home might be it would need to be part of a neutral foundation with assurances that the project is autonomous and governed by its contributors.

As a side benefit, this could get us out of the problem we've struggled with regarding "picking winners." We can extend mentorship to projects the Node.js ecosystem depends on without hosting them forever in "The Node.js TSC" which has an implicit endorsement.

mikeal:
Closing in favor of #59 for further discussion.

PR #59, opened 2016-03-08
Flatten project, scope TSC, big changes to structure.

mikeal:
Based on the recent TSC meeting and some of the feedback I've gotten on recent issues I've logged I created this initial draft of a new structure.

This does a few critical things:

* Sets the boundary for "long term project hosting" under the TSC to projects that are connected to Core development. However, it makes it clear that projects that are connected to Core development may stretch far beyond the needs of Core (evangelism, website, documentation).
* Renames the "Incubator" to "Mentorship" in order to reduce many false assumptions from Apache's Incubator. This still reserves the ability for the TSC to intervene in projects outside of the above scope, like Express, without assuming the TSC will host them long term. The issue of where to host these projects long term is a question for another day -- there are many options we can explore.
* Flattens all projects and Working Groups under the TSC. The line between what should and should not be under the CTC was always a bit difficult to draw. Flattening makes autonomy more clear and allows the TSC to focus on improving communication and collaboration between the different parts (which the CTC has never done much of because it's so focused on the needs of direct Core development).

There's some very good things this accomplishes.
*  It solves the "blessing projects" problem. By not hosting projects next to Core we can avoid making them "official."
* It makes the TSC focus on some core problems we're experiencing at scale.
 * Automating reporting, communication and transparency for projects and working groups.
 * Creating and promoting contribution and governance policies for use both within the TSC and **in the ecosystem**.
 * Creating measurements and tools for evaluating project health. First for use within the Foundation but conceptually for use outside it later on.

The good thing about flattening this stuff and staying away from a hierarchy is that we do a better and more explicit job of groups being service providers. This allows group to be more effective in this role within the project but also with the broader community.

What is not yet resolved is the question of who is on the TSC. Proportional representation is difficult, anything as simple as "one rep per project" ignores the disparity in contributors and scope between projects.

This document also isn't enough, there's a lot of other process points in other now irrelevant documents.

And finally, because this is a huge change that effects every group and project it's going to need to reach a consensus among all the sub projects. @nodejs/tsc @nodejs/ctc

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,30 +24,24 @@ Adding a new project or working group can be done through a pull request. Since
 each group is autonomous any new group taking on responsibility from another
 group would need that group's approval.
 
+Any project or working group can create teams at their discretion. Any org member
+can create a new repo for organizing this kind of effort without a charter or other
+process. Examples of Core teams are: Long Term Support, Cryto, Roadmap, Benchmark,
+Streams, Testing.
+
 ## Projects
 
 * Core
 * Website
-* Streams
-* Docker
 * nan
 
 ## Working Groups
 
 * Inclusivity
 * Build
-* Tracing
 * i18n
 * Evangelism
-* Roadmap
-* Addon API
-* Benchmarking
-* Post-mortem
-* Intl
-* HTTP
 * Documentation
-* Promises
-* Security
 
 ## Mentorship Program
 
--- a/README.md
+++ b/README.md
@@ -1,62 +1,66 @@
 # The Node.js Foundation TSC
 
-The Node.js Foundation Technical Steering Committee is the technical governing body of the Node.js Foundation. It admits and oversees all top-level Projects in the Node.js Foundation. It also elects a representative to the Node.js Foundation Board of Directors.
+This is the primary communication and governance hub of the Node.js
+Foundation's Technical Steering Committee. Its goal are:
 
-For more details read the [TSC Charter](https://github.com/nodejs/TSC/blob/master/TSC-Charter.md) adopted by the Node.js Foundation Board of Directors on June 17th 2015.
+* Maintaining, releasing and improving a healthy Node.js Core project.
+* Fostering and encouraging a healthy Node.js open source ecosystem.
 
-If your project is interested in joining the Node.js Foundation please read the [Project Lifecycle.md](./Project-Lifecycle.md) documentation.
+# TSC
+
+The TSC has many working groups and projects. Each group and project
+has a connection to the development of Node.js Core but their work and
+responsibilities often extend far beyond Core and potentially even Node.js.
+
+The TSC also leads several efforts to improve the health and sustainability
+of the ecosystem but it not a long term host of ecosystem projects which are
+not integrated into Core development.
+
+Each effort, project, and working group is autonomous. The scope of work and 
+autonomy is described in their charter. The TSC works to increase communication 
+and collaboration between these parts.
+
+Adding a new project or working group can be done through a pull request. Since
+each group is autonomous any new group taking on responsibility from another
+group would need that group's approval.
+
+## Projects
+
+* Core
+* Website
+* Streams
+* Docker
+* nan
+
+## Working Groups
+
+* Inclusivity
+* Build
+* Tracing
+* i18n
+* Evangelism
+* Roadmap
+* Addon API
+* Benchmarking
+* Post-mortem
+* Intl
+* HTTP
+* Documentation
+* Promises
+* Security
+
+## Mentorship Program
+
+The TSC administers a mentorship program for critical projects in the ecosystem.
+Projects are assigned mentors from the TSC to adapt modern best practices to the 
+specific needs of each project.
+
+* libuv
+* Express
+
 
 ## TSC Members
 
-TSC members are responsible for top level technical community concerns. The role is 
-mostly administrative and is responsible for admitting new Top Level Projects, Top Level
-Working Groups, and advocating for any needs in the technical side of the foundation to
-the Node.js Foundation Board of Directors.
-
-TSC members can nominate new members at any time. Candidates for membership tend to be people
-who have a competancy for community management and a high tolerance and patience for process
-minutiae as the TSC delegates most of its responsibilities to other projects and working groups.
-
-Every Top Level Project not currently incubating can appoint someone to the TSC who they elect
-at their own discretion.
-
-## Top-Level WGs and TLPs
-
-* [Working Groups](WORKING_GROUPS.md)
-* Mentors
-
-    Project mentorship is not a technical role. In fact, mentors are
-    discouraged from giving technical advice to projects. Instead, the
-    purpose of mentorship is to encourage and improve a projects ability
-    to be participatory, transparent, and effective. Mentors are there to
-    help projects adopt and iterate on policies and processes that achieve
-    these goals and eventually allow them to graduate the incubation phase.
-
-      * Mikeal Rogers (@mikeal)
-* Top-Level Projects
- * Core TLP
-  * Core WGs (streams, http, Intl)
-
-## Policy Change Proposal Process
-
-The Node.js TSC is chartered to oversee the technical governance of all Top
-Level Projects and Working Groups under the Node.js Foundation. The TSC
-establishes the default governance, conduct, and licensing policies for all Top
-Level Projects. Top Level Projects and Working Groups have broad powers of
-self-governance.
-
-To propose a change or addition to policies or processes that are intended to
-cover all Top Level Projects and Working Groups in the foundation, a PR should
-be opened in the `nodejs/TSC` repository.
-
-The pull request can be labeled `tsc-agenda` to request that it be put on the
-agenda for the next TSC meeting.
-
-The Node.js Foundation Board of Directors retains certain rights (especially
-legal considerations). If the TSC endorses a proposal, they will escalate to the
-Node.js Foundation Board of Directors when required to do so.
-
-In some cases, existing individual groups have the right to refuse changes to
-their charters. The TSC can not mandate existing working groups alter their
-charters. If such a situation arises, the TSC may decide to revoke the group's
-charter.
+TSC members are responsible for administration and communitication. The role is 
+mostly administrative and is responsible for admitting new projects, working groups 
+and representing them to  Node.js Foundation Board of Directors.
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 # The Node.js Foundation TSC
 
 This is the primary communication and governance hub of the Node.js
-Foundation's Technical Steering Committee. Its goal are:
+Foundation's Technical Steering Committee. Its goals are:
 
 * Maintaining, releasing and improving a healthy Node.js Core project.
 * Fostering and encouraging a healthy Node.js open source ecosystem.
@@ -13,12 +13,12 @@ has a connection to the development of Node.js Core but their work and
 responsibilities often extend far beyond Core and potentially even Node.js.
 
 The TSC also leads several efforts to improve the health and sustainability
-of the ecosystem but it not a long term host of ecosystem projects which are
+of the ecosystem but it is not a long term host of ecosystem projects which are
 not integrated into Core development.
 
-Each effort, project, and working group is autonomous. The scope of work and 
+Each project and working group is autonomous. The scope of work and 
 autonomy is described in their charter. The TSC works to increase communication 
-and collaboration between these parts.
+and collaboration between all these parts.
 
 Adding a new project or working group can be done through a pull request. Since
 each group is autonomous any new group taking on responsibility from another
@@ -57,7 +57,6 @@ specific needs of each project.
 
 * libuv
 * Express
-
 
 ## TSC Members
 


williamkapke:
Awesome- now I know where to find a list of all the committees!

README.md
```diff
@@ -1,62 +1,65 @@
...
-The Node.js Foundation Technical Steering Committee is the technical governing body of the Node.js Foundation. It admits and oversees all top-level Projects in the Node.js Foundation. It also elects a representative to the Node.js Foundation Board of Directors.
+This is the primary communication and governance hub of the Node.js
+Foundation's Technical Steering Committee. Its goals are:
 
-For more details read the [TSC Charter](https://github.com/nodejs/TSC/blob/master/TSC-Charter.md) adopted by the Node.js Foundation Board of Directors on June 17th 2015.
+* Maintaining, releasing and improving a healthy Node.js Core project.
+* Fostering and encouraging a healthy Node.js open source ecosystem.
 
-If your project is interested in joining the Node.js Foundation please read the [Project Lifecycle.md](./Project-Lifecycle.md) documentation.
+# TSC
```
williamkapke: There is already a `#  The Node.js Foundation TSC` above... this could probably be left out.

README.md
```diff
@@ -1,62 +1,65 @@
 # The Node.js Foundation TSC
 
-The Node.js Foundation Technical Steering Committee is the technical governing body of the Node.js Foundation. It admits and oversees all top-level Projects in the Node.js Foundation. It also elects a representative to the Node.js Foundation Board of Directors.
+This is the primary communication and governance hub of the Node.js
+Foundation's Technical Steering Committee. Its goals are:
 
-For more details read the [TSC Charter](https://github.com/nodejs/TSC/blob/master/TSC-Charter.md) adopted by the Node.js Foundation Board of Directors on June 17th 2015.
+* Maintaining, releasing and improving a healthy Node.js Core project.
+* Fostering and encouraging a healthy Node.js open source ecosystem.
```
williamkapke: `healthy` is out of place. Consider rewording.

README.md
```diff
@@ -1,62 +1,65 @@
...
 
-## TSC Members
-
-TSC members are responsible for top level technical community concerns. The role is 
-mostly administrative and is responsible for admitting new Top Level Projects, Top Level
-Working Groups, and advocating for any needs in the technical side of the foundation to
-the Node.js Foundation Board of Directors.
+The TSC has many working groups and projects. Each group and project
+has a connection to the development of Node.js Core but their work and
+responsibilities often extend far beyond Core and potentially even Node.js.
```
williamkapke: maybe just:
> ...extend far beyond <del>Core and potentially even Node.js</del>.

README.md
```diff
@@ -1,62 +1,65 @@
...
 
-TSC members can nominate new members at any time. Candidates for membership tend to be people
-who have a competancy for community management and a high tolerance and patience for process
-minutiae as the TSC delegates most of its responsibilities to other projects and working groups.
+The TSC also leads several efforts to improve the health and sustainability
+of the ecosystem but it is not a long term host of ecosystem projects which are
+not integrated into Core development.
```
williamkapke: > ... into Node.js Core development
or
> ... into Core's development
or..?
> ... into core development

README.md
```diff
@@ -1,62 +1,65 @@
...
-Node.js Foundation Board of Directors when required to do so.
+## TSC Members
 
-In some cases, existing individual groups have the right to refuse changes to
-their charters. The TSC can not mandate existing working groups alter their
-charters. If such a situation arises, the TSC may decide to revoke the group's
-charter.
+TSC members are responsible for administration and communitication. The role is 
+mostly administrative and is responsible for admitting new projects, working groups 
+and representing them to  Node.js Foundation Board of Directors.
```
williamkapke: Would/should a current list go here? This is what I would expect/hope to find here.

mikeal: We should figure out any alterations before this is merged.

README.md
```diff
@@ -1,62 +1,65 @@
...
-The Node.js Foundation Board of Directors retains certain rights (especially
-legal considerations). If the TSC endorses a proposal, they will escalate to the
-Node.js Foundation Board of Directors when required to do so.
+## TSC Members
 
-In some cases, existing individual groups have the right to refuse changes to
-their charters. The TSC can not mandate existing working groups alter their
-charters. If such a situation arises, the TSC may decide to revoke the group's
-charter.
+TSC members are responsible for administration and communitication. The role is
```
jasnell: s/communitication/communication

README.md
```diff
@@ -1,62 +1,65 @@
...
+* nan
 
-      * Mikeal Rogers (@mikeal)
-* Top-Level Projects
- * Core TLP
-  * Core WGs (streams, http, Intl)
+## Working Groups
 
-## Policy Change Proposal Process
+* Inclusivity
```
jasnell: It should be noted that the WG's below are chartered by the TC and not the TSC.

mikeal: I'm proposing changing where they get chartered in the future so it would be confusing to separate them.

README.md
evanlucas: two spaces between `to` and `Node.js`

README.md
nebrius: Will the other groups need to be rechartered and/or re-approved by the TSC, or will they be sort of "grandfathered" in? If it's the former, maybe we could denote with an asterisk here that they are pending rechartering?

mikeal: I don't see why they would need to be re-chartered unless we think their scope could be miss-interpreted as broader now that they are at a higher level. This would however need the approval of CTC.

README.md
nebrius: Are you saying that `healthy` is out of place in both line 6 and 7, or just one of the lines? Can you explain why healthy is out of place? I think it's a good addition myself, especially on line 7, but I'd like to hear your reasoning.

README.md
nebrius: I can think of a handful of reasons we might want to consider rechartering, including misinterpreting change of scope. Is this PR the best place to discuss that, or should we open a new issue and discuss there so we don't derail this PR?

mikeal: Hrm... this is tricky. We should probably move the WORKING_GROUPS.md from nodejs/node here and commit it as is in this PR and then open another one over it for the re-scoping so that we can better track the changes. That should work fine though, there's no point in writing new charters if this PR doesn't pass.

README.md
williamkapke: Ugh. I hate to dive in to such a nit-pick, sorry in advance to everyone....

Line 6 is the one that is a bit off. Adding `health` makes it the subject of the actions (which I believe is correct & important). You can `maintain` and `improve` health- but not `release` it. I believe `release` is intended to have Node.js Core (code) as the subject- which obviously you can release code. Careful guidance & oversight of the releases of Node.js Core releases will maintain and improve it. I'm sure everyone will 'get the idea' and probably just read past- so that's why I just said "consider rewording."

mikeal: Ya, this can be worded better.

nebrius: That makes sense, thanks @williamkapke, and yeah I agree now it should be reworded too.

README.md
nebrius: Sounds good to me

jasnell: hmmm... I'm not sure changing where those are chartered is a good thing. The core TC needs to be able to delegate responsibility on it's own independently of the TSC... unless you're suggesting flattening the TC and TSC back into one?

mikeal: If the TC is delegating then the TC *and* TSC will approve the charter. See
the language regarding transfer of responsibilities. The point is that once
the TC delegates something to a subgroup the TSC will be following up to
make sure they communicate and document well, not the TC because we know
from experience they don't ;)

For a long time we basically relied on people who happened to be in all
these sub groups also being in the TC but that isn't scaling.

joshgav:
Can we more explicitly define "Project" and "Working Group"? A quick read of [Project-Lifecycle.md](https://github.com/nodejs/TSC/blob/master/Project-Lifecycle.md) indicates that a Project can have WGs, while WGs cannot have (sub-)WGs. Are there other differences?

We could rationalize all top-level things as Working Groups, drop the "Project" moniker, and then provide optional guidance for subcommittees for WGs which want them.

Re current CTC WGs, perhaps they should be evaluated one by one and either become a proper top-level WG or remain a subcommittee of Core. Some of them could perhaps be closed or consolidated in this process - for example post-mortem and tracing, intl and i18n, Addon API and nan, etc.

It would be helpful if all of this was documented in one place, probably this nodejs/TSC repo, rather than duplicated/shared with nodejs/node. The nodejs.org website should also be synced with info in this repo; many people start there.

Thanks!

mikeal:
@joshgav first off, Project Lifecycle will need to be completely re-written if this is passed.

The main difference between a project and a working group is that a project has a definable code base it works on and a working group is broader in scope and probably only using a repo for communication and tracking purposes.

README.md
jasnell: Still not convinced. The TC for core ought to be able to spin out it's own WG's to focus on core stuff in a manner that is entirely independent of what is happening on the TSC side.

Just carrying this through a bit... would you imagine that any WG's spun up by the Express TC would *also* fall under the TSC using this plan? So hypothetically if we ended up with separate "Core LTS" and "Express LTS" WG's those would both exist under the TSC? Or is there something else that you had in mind that I'm just not groking from this?

joshgav:
@mikeal From your explanation I see why a WG isn't necessarily a project; but perhaps a project is by definition a WG? I'm trying to rationalize this as much for myself as for others :).

The Node project could be a collection of WGs governed by the TSC. WGs could include any or all of
- code,
- an issue tracker/discussion forum,
- a chief Technical Committee
- subcomittees

README.md
mikeal: Express is a special snowflake because it's in the mentorship program with the goal of being its own fully contained project in the future, potentially hosted outside the TSC, so no this doesn't hold for Express.

You make a good point though that a subgroup should be allowed to delegate of its own will without approval from the TSC. My only thought was the the TSC would be responsible for keeping communication clear, so having a hand in making sure a charter is well defined would be important to them, but I don't see a need for the TSC to *agree* to every charter, just that they will probably want to be aware and possibly involved in the creation of any charters.

With regard to LTS needing to be *under* a TC, I'm not sure that entirely holds true. The LTS WG is actually still unchartered, it *technically* has no authority, but the TC defers to it often. The LTS WG also doesn't define its own release team, that's all integrated into Core, the WG is actually similar to the Streams WG in that it does work but the final decision about that work being merged/released falls under the control of the CTC -- it just so happens that LTS WG members are committers and CTC members so this is all very well integrated, but the CTC actually haven't officially delegated the committing and releasing to that WG.

joshgav:
For the sake of discussion, here's an idea of what the collection of WGs and subcommittees might look like:

- **Core** -> Build, Testing, Benchmarking, Streams, HTTP
- **Futures** -> Roadmap, Promises, Docker
- **Diagnostics** -> Tracing, Post-Mortem
- **Modules** -> nan, Addon API, npm
- **Security**
- **Docs**
- **i18n**
- **Evangelism** -> Website
- **Inclusivity**

Fishrock123:
@joshgav I don't think that would work well.

> - **Futures** -> Roadmap, Promises, Docker

Roadmap-related things are nothing if not associated to core.

> - **Diagnostics** -> Tracing, Post-Mortem
- **Modules** -> nan, Addon API, npm

Likewise

> - **Evangelism** -> Website

Those two definitely do not directly overlap like that.

mikeal:
I think we're missing the point a little.

The basic idea here is to *limit* the scope of the TSC (with the exception of mentoring other projects) to coordinating sub-groups that are *integrated into Core development in some way.* Almost all of these groups have a broader set of responsibilities beyond Core development but some part of them needs to be integrated nonetheless.

Consolidating groups is a poor plan for the simple reason that many of these groups are already huge and consolidating them would be a poor plan being that some of them are literally the result of them delegating responsibilities in order to scale.

Measuring the links between groups and their connections to each other and Core is incredibly difficult and actually changes over time. I'm convinced it's actually impossible to maintain a description of these connections, we attempted to once six months ago and ended up with a crazy web which has only gotten crazier since then.

Instead, all I think anyone can expect the TSC and these groups to do is find better ways of communicating the work they are already doing. These groups operate in the open but the volume of work is impossible to understand without better reporting and communication. If we have better communication all the people working in these groups have the ability to understand *as much as they need to understand* about each sub-group.

This was working a lot better and tighter when we could rely on a lot of common members across groups, that doesn't seem to work anymore, and even people who spend a considerable amount of time trying to understand what is going on across the org can't do it. If people like me and @rvagg can't figure this out then we have to assume that absolutely nobody in the community has an understanding of this.

nebrius:
> From your explanation I see why a WG isn't necessarily a project; but perhaps a project is by definition a WG? I'm trying to rationalize this as much for myself as for others :).

This is an imperfect and unofficial definition, to be sure, but I think of projects as producing a software product that can be used by developers inside and outside of the Node.js project. Conversely, I view WGs as organizations where people discuss topics related to one or more projects in order to influence how projects create their products.

mikeal:
Feedback from @jasnell in CTC meeting: Several of these are subsets of Core development and shouldn't be put under the TSC directly.

Response from @mikeal was that those groups look more like *teams* that Working Groups and that these were chartered as WGs because there's a perception that if you want to use another repo to organize work it needs to be a WG.

Positive response from @jasnell that we should maybe just make these teams and be clearer that nodejs org members can use multiple repos freely to coordinate.

README.md
rvagg: My original wording for that point was simply "* Maintaining the health of the core Node.js project."

joshgav:
A thought: What if the TSC and CTC reunited into a chief TC (TSC), and the role defined for the TSC now became a Project Management WG under the chief TC?

Since @mikeal [proposes (comment)](https://github.com/nodejs/TSC/pull/59#issue-139137860) that we define the domain of the TSC as projects relevant to Core, the domain of the CTC and TSC mostly overlaps; but they serve different roles in this domain. At Microsoft we might call these "dev" and "program management (PM)." CTC members need to focus on engineering, TSC on cross-team coordination and project management. A reunited CTC/TSC would manage all aspects of core, delegating PM duties to the PM workgroup. All other workgroups would then report to the united TSC and CTC.

Possibly, rather than a sub-WG, the chief TC could be composed of both project management people (TSC) and core devs (CTC).

Thoughts?

Fishrock123:
Just as a note, someone in Q/A for the TSC meeting today brought up that why this is even necessary may not be apparent without actually listening to previous TSC meetings. We should probably express somewhere in written form here on our concerns about adding new projects.

mikeal:
I made some updates to this based on the feedback I got last week.

* Reduce the number of projects and working groups. Many of these were just teams that had a repo they were using for organization.
* Reduce process for creating teams. It was somewhat telling that one of the most active "working groups" was never even chartered (LTS). Rather than run through the whole chartering process this just lets project create teams and repos at will without extraneous process. The only reason to charter a WG or new project is to actually delegate and remove responsibility from another group.

Fishrock123:
> I am that someone. Does it anonymize us on YouTube? Maybe I'll go back to using #node-dev then :/

@williamkapke No, but it's not always apparent what usernames are across services or if people want to be mentioned. :)

mikeal:
Here's a basic drawing of the latest proposal.

Obviously libuv and express aren't "teams" but the rest is accurate.

screen shot 2016-03-15 at 5 20 07 pm

williamkapke:
@mikeal so the CTC's oversight fits in where "Core" is?

mikeal:
@williamkapke correct.

orangemocha:
One concern I have with this PR is that it bundles two sets of important changes:

1. The decision to focus the TSC on Core
2. The organization restructuring to accomplish that mission

The new organizational structure makes sense to me, *once we have decided to focus on Core*. But the fundamental decision of what the TSC mission should be is something that we have been trying to figure out for quite some time now, and that in my opinion still needs to be resolved first. (I am also noting that this is a 180-degree turn from the proposal of just a few weeks ago).

The end result might be the same, but I would prefer a separate PR about the TSC mission, so that we can have a focused discussion about it and finally reach a consensus.

mikeal:
> (I am also noting that this is a 180-degree turn from the proposal of just a few weeks ago).

Sure, but that proposal wasn't accepted because nobody was comfortable with that direction :)

> One concern I have with this PR is that it bundles two sets of important changes

The TSC can discuss each of these points in isolation but I don't see how it benefits the discussion to bring the entire proposal back to first principals because most people have a hard time visualizing the actual impact of the first discussion without the second point -- that's why they are bundled. 

In particular, both decisions have an impact on the CTC and without the structural proposal I don't think people who aren't active in the TSC meetings will be able to understand the impact.

mikeal:
> The decision to focus the TSC on Core

So, maybe the language needs to be clearer, but this doesn't focus the TSC "on Core" but on integrating various groups into Core development. Perhaps that can be made clearer.

joshgav:
Since the task of Build is to build Core, should it be part of Core? Or what's the reasoning behind it being separate?

nan seems to be part of a larger domain for core extensibility and the module subsystem. For example, that domain might include [nodejs/api#10](https://github.com/nodejs/api/issues/10) and [nodejs/node-eps#3](https://github.com/nodejs/node-eps/pull/3).

mikeal:
@joshgav roughly 90% of what Build *builds* is Node.js. They also handle the hosting the website, micro-sites, etc. They also have build and test resource for nan and libuv. The Build WG manages a lot of infrastructure and operates essentially independently already so it makes sense to retain this level of autonomy.

> nan seems to be part of a larger domain for core extensibility and the module subsystem. For example, that domain might include nodejs/api#10 and nodejs/node-eps#3.

Even if an addon api or new core native API were to happen we would still need to maintain nan, it's highly depended on and needs to remain independent. It also doesn't make sense to put alternatives to nan *under* nan so those should remain in Core.

joshgav:
@mikeal 
> It also doesn't make sense to put alternatives to nan under nan so those should remain in Core.

Actually I was thinking they all belong together under `extensibility`, `modules` or some such. Which will raise the question of whether that should be under Core :)

nebrius:
Do I remember correctly that intl does website work as well? If so, should they be top level?

nebrius:
I could get behind the idea of a "modules" or "native modules" top level group, of which NAN is one team under them.

mikeal:
> Do I remember correctly that intl does website work as well? If so, should they be top level?

There's i18n, which is the localization communities, and Intl which is a group working on the Intl support in core :)

README.md
```diff
@@ -1,62 +1,59 @@
...
-at their own discretion.
+Adding a new project or working group can be done through a pull request. Since
+each group is autonomous any new group taking on responsibility from another
+group would need that group's approval.
 
-## Top-Level WGs and TLPs
+Any project or working group can create teams at their discretion. Any org member
+can create a new repo for organizing this kind of effort without a charter or other
+process. Examples of Core teams are: Long Term Support, Cryto, Roadmap, Benchmark,
+Streams, Testing.
```
Fishrock123: needs some definite of what responsibility teams are allowed to take and when

README.md
```diff
@@ -1,62 +1,59 @@
...
 
-* [Working Groups](WORKING_GROUPS.md)
-* Mentors
+## Projects
```
Fishrock123: what is the different between TLP and WG?

Fishrock123:
Who gets TSC seats from this new structure then, if anyone?

mikeal:
> Who gets TSC seats from this new structure then, if anyone?

I left that for the TSC to figure out after you've agreed to the scope. The scope sort of dictates the kind of representation you'd want to see.

> WG vs. Project

No difference really, WG is just an easier term to refer to a slate of responsibility when it's larger than a code base.

williamkapke:
@mikeal, I think the drawing you created provides tremendous context for the proposed change(s). I think it would really help the discussion to have a drawing for the EXISTING structure. Basically, a before and after.

I think it would also serve this thread well to edit the OP to include the chart(s) and give that visual right away.

README.md
```diff
@@ -1,62 +1,59 @@
...
 
-    Project mentorship is not a technical role. In fact, mentors are
-    discouraged from giving technical advice to projects. Instead, the
-    purpose of mentorship is to encourage and improve a projects ability
-    to be participatory, transparent, and effective. Mentors are there to
-    help projects adopt and iterate on policies and processes that achieve
-    these goals and eventually allow them to graduate the incubation phase.
+* Core
+* Website
+* nan
```
rvagg: NAN

Fishrock123:
I think I've brought this up in the meetings but:

1. While I'm not against restructuring if necessary, I'd rather not do so again so soon.
2. I'm not sure how putting more communicative work on the TSC will actually improve org-wide communication in the short term.
3. Having the possibility to still take projects on in a longer term basis _may_ still be desired, but definitely should not be the default or something you can just apply for upfront.

rvagg:
While I don't mind the majority of the way this is structured now I'm pretty sure there isn't a mood for doing yet another big pivot. The question of scope is an important one but I'd much prefer to frame in terms of a consultation within the Foundation. The TSC has an obvious role in core, we have (and should have), a lot of latitude to do what we think is needed there and we have the trust of the board on that. Beyond core, as @Fishrock123 says, there may be situations where limiting our scope hinders our ability to do real good for the Node ecosystem. _But_, I think it's completely appropriate for those actions to trigger a conversation with the Board (in particular, but also the executive) regarding broader strategy.

This is the theme I'm going to be pushing next week at the in-person Board meeting. I don't think it's in our interests to act on this here before that discussion and limit our options prematurely. I expect that the Board will be asking us to limit our options somewhat because I don't think they have any stomach for the Foundation to become the next Apache Foundation, but I agree with that and would like to encourage us to think about expansion in terms of strategy where one arm of the Foundation can propose to the rest a course of action but which requires agreement from the other arms to execute on that.

All centered around a "mission", which is kind of terrible in the by-laws, which is why I'm proposing:

**The Node.js Foundation exists to increase the adoption of Node.js.**
This goal is achieved by:
1. Maintaining the health of the core Node.js project
2. Fostering and encouraging a healthy Node.js open source ecosystem
3. Fostering and encouraging a healthy commercial ecosystem of training, products and services for Node.js

mikeal:
@rvagg can you work that up into a PR or a new Issue that can be tagged for the agenda tomorrow and potentially passed by the TSC?

mikeal:
I'm going to close this.

I wrote up what appeared to be the consensus a few meetings back. It is now clear that it is not the consensus, and that there may not be a consensus to this level of detail.

I'll leave it to Rod to write up a less detailed statement that might reach a consensus.

williamkapke:
@mikeal you closed #54 and #58 in favor of this thread... which is now closed too. Should you reopen them?

@rvagg:
> While I don't mind the majority of the way this is structured now I'm pretty sure there isn't a mood for doing yet another big pivot.

What _is_ the _current_ structure of the Technical arm of the Foundation? I, as an outsider that has followed pretty closely over the years, can't figure it out.

mikeal:
@williamkapke nah, leave them closed, we should defer these topics to whatever Rod ends up logging.

PR #72, opened 2016-03-22
MEETINGS.md to provide quick access to info

williamkapke:
Perhaps if meeting info is surfaced there will be more participation! 🎉

### DIFF
--- a/MEETINGS.md
+++ b/MEETINGS.md
@@ -1,0 +1,25 @@
+# TSC Meetings
+
+The TSC has weekly meetings to discuss issues labeled with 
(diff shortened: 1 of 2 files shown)


mikeal:
Should we point people at a single link for a tag rather than every individual meeting? If we did then we wouldn't have to patch this file every week, and this could maybe just go in the README for better visibility.

Also, the TSC was doing meetings every other week for a while and might return to that at some point.

williamkapke:
I'm hoping this can be integrated into [the bot](https://github.com/nodejs-github-bot/github-bot) so the updates are done automatically once the issue is closed.

Also, I agree- this would be better in the README.

PR #80, opened 2016-03-25
List of TSC Members with more flare ✨ ✨

williamkapke:
Here's an alternate option (to #1) for the list of TSC members.
I wanted something that adds color to the page. Not sure how people feel about markdown this ugly.

I ordered them by first name but put @rvagg at the top with special attribute to his _special_ role.


williamkapke:
Ping for review. Or maybe approve #1... either would be awesome!

rvagg:
lgtm

PR #86, opened 2016-04-02
Added link about personal identity

nebrius:
While reviewing https://github.com/nodejs/nodejs.org/pull/623, I noticed that the list of identity traits was removed from the list of responsibilities for the Inclusivity WG section from [TSC/WORKING_GROUPS.md](https://github.com/nodejs/TSC/blob/master/WORKING_GROUPS.md) that is present in [inclusivity/README.md](https://github.com/nodejs/inclusivity/blob/master/README.md).

I think this list of identities is important for understanding what this responsibility means, but at the same time I understand how much length it adds to the document and why it was removed.

As a compromise, I added a link to the Inclusivity WG's section on responsibilities so that people can discover the list and the full meaning of this responsibility without bloating the length of this document.

### DIFF
--- a/WORKING_GROUPS.md
+++ b/WORKING_GROUPS.md
@@ -40,11 +40,11 @@ for the Node.js project:
 Its responsibilites are:
 
 * Foster a welcoming environment that ensures participants are valued and can
-feel confident contributing or joining discussions, regardless of any aspect of 
-their identity.
+feel confident contributing or joining discussions, regardless of any [aspect of
+their identity](https://github.com/nodejs/inclusivity/#list-of-responsibilities).
 * Proactively seek and propose concrete steps the project can take to increase
 inclusivity.
-* Serve as a resource for the development and enforcement of workflows that 
+* Serve as a resource for the development and enforcement of workflows that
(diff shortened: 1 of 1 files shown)


WORKING_GROUPS.md
```diff
@@ -40,11 +40,11 @@ for the Node.js project:
 Its responsibilites are:
```
williamkapke: Can you add a comment here like:
```
<!--
This information should mostly mirror:
https://github.com/nodejs/inclusivity/blob/master/README.md#list-of-responsibilities
-->
```

Also, consider doing a PR to [inclusivity/README.md](https://github.com/nodejs/inclusivity/blob/master/README.md) with comments to help any possible future changes get propagated to the mirrors.

williamkapke:
Seems like a great compromise! LGTM with the nit about adding a comment.

WORKING_GROUPS.md
nebrius: Good call, will do.

nebrius:
I added links to the responsibilities and purpose, PTAL

williamkapke:
LGTM!

PR #102, opened 2016-05-18
doc: update the moderation policy

jasnell:
Fill in some ambiguous gaps in the moderation policy

/cc @nodejs/tsc

### DIFF
--- a/Moderation-Policy.md
+++ b/Moderation-Policy.md
@@ -31,10 +31,11 @@ Any alternative Moderation Policy used for a given repository must be included i
 
 ### Terms
 
-* *Collaborator* refers to any individual with configured access permissions to any Node.js GitHub organization repository. See [GitHub's access permissions documentation](https://help.github.com/articles/what-are-the-different-access-permissions/) for more information.
+* *Collaborator* refers to any individual with configured write (commit) permissions to any Node.js GitHub organization repository *other than the Moderation Repository*. See [GitHub's access permissions documentation](https://help.github.com/articles/what-are-the-different-access-permissions/) for more information.
 * *TSC* refers to the [Node.js Technical Steering Committee](https://github.com/nodejs/node#tsc-technical-steering-committee).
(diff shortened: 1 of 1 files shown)


Moderation-Policy.md
```diff
@@ -31,10 +31,11 @@ Any alternative Moderation Policy used for a given repository must be included i
 
 ### Terms
 
-* *Collaborator* refers to any individual with configured access permissions to any Node.js GitHub organization repository. See [GitHub's access permissions documentation](https://help.github.com/articles/what-are-the-different-access-permissions/) for more information.
+* *Collaborator* refers to any individual with configured write (commit) permissions to any Node.js GitHub organization repository *other than the Moderation Repository*. See [GitHub's access permissions documentation](https://help.github.com/articles/what-are-the-different-access-permissions/) for more information.
```
nebrius: Does everyone in node/membership have write access specifically to the moderation repo? I could be wrong, but I thought I remembered that only a subset of people have write access.

jasnell: Every Collaborator should have at least access to view the Moderation Repository and to participate in discussions.

Charlotteis:
This certainly clarifies a few things for me. Thanks jasnell. I particularly like the `write (commit)` addition to the Collaborator definition as this is a really good way to describe it.

bnoordhuis:
LGTM

Moderation-Policy.md
nebrius: I could be wrong, but I thought that being a member of the org gave read access, but not write.

On further reflection though, if this is the case, it doesn't make the wording here wrong. Absolute worst case it's a bit superfluous, and best case it makes an important distinction, so there's no downside. I'm ok with this.

nebrius:
LGTM

mhdawson:
LGTM

TheAlphaNerd:
@jasnell can this land?

jasnell:
Yep

williamkapke:
Lotsa LGTMs here. Merging.

Issue #89, opened 2016-04-15
Define difference between collaborator and committer


phillipj:
I'm working on a talk about contributing open source to the Node.js project, and noticed both **collaborator** and **committer** mentioned in this repo. In the [current contributing guide](https://github.com/nodejs/TSC/blob/master/BasePolicies/CONTRIBUTING.md), collaborator isn't even mentioned. In fact that's the only document I've seen which use the term committer.

It would have been helpful with an explanation between the two terms. The only hint I've got so far is a comment reply in the PR creating the contributing guide: https://github.com/nodejs/TSC/pull/40#discussion_r51919768

/cc @mikeal

Fishrock123:
Same thing at the current time.

We should move towards using the term committer instead for various reasons though, but that will take some time I think. (And may be forgotten..)

phillipj:
> We should move towards using the term committer instead for various reasons though

Thanks, that at least gives me an answer about what term I should use.

@Fishrock123 would it be appropriate to s/collaborator/committer in `README.md` in the repos we've got under the nodejs org?

Trott:
@phillipj I would be cautious about that. Especially in the `node` repo, the term `collaborator` was no doubt chosen very deliberately over the term `committer`, probably for some (perhaps all) of the following reasons:

* emphasize consensus-driven decision making
* de-emphasize hierarchy (especially BDFL model)
* emphasize that code is not the only valued form of contribution
* de-emphasize the role as one that gets bestowed privileges and emphasize it as one that welcomes someone to a community

(No opinion from me on the TSC repo, though. I'm just chiming in because you asked about other repos, and I've got an opinion.)

Fishrock123:
cc @mikeal

williamkapke:
This may fall under "teams have autonomy" and can call them whatever they want. [Which I don't think provides a good/cohesive environment for newcomers](https://github.com/nodejs/TSC/issues/81#issuecomment-203257489) (just my opinion though).

Also related: https://github.com/nodejs/collaboration

phillipj:
@Trott thanks for your input. You understood correctly, one of the readmes I was thinking about was absolutely node core as it probably gets most attention, and therefore would have helped the most to get the committer term out there. Too bad it wasn't that simple after all.

IMO using these two terms easily leads to confusion. An explicit explanation of what separates the two would be most valuable.

fhemberger:
[...]

PR #104, opened 2016-05-22
Add term "collaborator" to CONTRIBUTING.md

phillipj:
I've previously opened issue https://github.com/nodejs/TSC/issues/89 to clarify the difference between "committer" and "collaborator". It's very confusing that `Baseline/CONTRIBUTING.md` doesn't even mention "collaborator", when that's clearly the term most feel comfortable using compared to "committer", ref comments in #89.

I couldn't really find a good definition for it my self which makes sense compared to "committer", any thoughts and suggestions are much appreciated. And if we can't really find a good definition for it, might I suggest we drop using "committer" and stick with "collaborator"?

Closes https://github.com/nodejs/TSC/issues/89

### DIFF
--- a/BasePolicies/CONTRIBUTING.md
+++ b/BasePolicies/CONTRIBUTING.md
@@ -19,6 +19,7 @@ and the Node.js project.
 
 * A **Contributor** is any individual creating or commenting on an issue or pull request.
 * A **Committer** is a subset of contributors who have been given write access to the repository.
+* A **Collaborator** is a committer, but may use more time labelling and closing issues, rather than commit code.
 * A **TC (Technical Committee)** is a group of committers representing the required technical 
 expertise to resolve rare disputes.
 


BasePolicies/CONTRIBUTING.md
```diff
@@ -19,6 +19,7 @@ and the Node.js project.
 
 * A **Contributor** is any individual creating or commenting on an issue or pull request.
 * A **Committer** is a subset of contributors who have been given write access to the repository.
+* A **Collaborator** is a committer, but may use more time tag and close issues, rather than commit code.
```
williamkapke: I think something went wrong with your the editing- the sentence didn't come out right. Maybe I'm just loosing my mind :/

phillipj: @williamkapke thanks! You're right, made some changes.

..and any suggestions would be much appreciated 😸

BasePolicies/CONTRIBUTING.md
```diff
@@ -19,6 +19,7 @@ and the Node.js project.
 
 * A **Contributor** is any individual creating or commenting on an issue or pull request.
 * A **Committer** is a subset of contributors who have been given write access to the repository.
+* A **Collaborator** is a committer, but may use more time labelling and closing issues, rather than commit code.
```
nebrius: I could be wrong (someone please clarify), but IIRC "collaborator" isn't quite an official term, even though we use it frequently.

Perhaps it might be useful to add some wording similar to

```
* Informally, a **Collaborator** is...
```

jasnell:
Closing due to lack of forward movement on this. Can reopen if necessary

Issue #69, opened 2016-03-17
Node.js Foundation Technical Steering Committee (TSC) Meeting 2016-03-17


rvagg:
## Time

**UTC Thu 17-Mar-2016 20:00**:

* San Francisco: Thu 17-Mar-2016 13:00
* New York: Thu 17-Mar-2016 16:00
* Amsterdam: Thu 17-Mar-2016 21:00
* Moscow: Thu 17-Mar-2016 23:00
* Sydney: Fri 18-Mar-2016 07:00
* Tokyo: Fri 18-Mar-2016 05:00

Or in your local time:
* http://www.timeanddate.com/worldclock/fixedtime.html?msg=Node.js+Foundation+Technical%20Steering%20Committee%20(TSC)+Meeting+2016-03-17&iso=20160317T20
* or http://www.wolframalpha.com/input/?i=8pm+UTC%2C+Mar+17%2C+2016+in+local+time

## Links

* **Minutes Google Doc**: <https://docs.google.com/document/d/1lFnhNaE2wK0FFyUfDz1LcaBiD3-rW_UKFjUQqqBIZOY>
* _Previous Minutes Google Doc: <https://docs.google.com/document/d/1Zt6EnkLghHUxqvn8RKz4ewy-uljIdhqU96KHv_uH6oc>_

## Agenda

Extracted from **tsc-agenda** labelled issues and pull requests from the **nodejs org** prior to the meeting.

### nodejs/TSC

* Node.js "Users" feedback [#66](https://github.com/nodejs/TSC/issues/66)
* Flatten project, scope TSC, big changes to structure. [#59](https://github.com/nodejs/TSC/pull/59)
* Add guidance to proposal section [#53](https://github.com/nodejs/TSC/pull/53)

## Invited

* @bnoordhuis (TSC)
* @chrisdickinson (TSC)
* @cjihrig (TSC)
* @Fishrock123 (TSC)
* @indutny (TSC)
* @jasnell (TSC)
* @misterdjules (TSC)
* @mikeal (observer)
* @mscdex (TSC)
* @orangemocha (TSC)
* @piscisaureus (TSC)
* @rvagg (TSC)
* @shigeki (TSC)
* @trevnorris (TSC)
* @nebrius

## Notes

The agenda comes from issues labelled with `tsc-agenda` across **all of the repositories in the nodejs org**. Please label any additional issues that should be on the agenda before the meeting starts.

## Joining the meeting

[...]

Issue #61, opened 2016-03-10
Node.js Foundation Technical Steering Committee (TSC) Meeting 2016-03-10


rvagg:
## Time

**UTC Thu 10-Mar-2016 20:00**:

* San Francisco: Thu 10-Mar-2016 12:00
* New York: Thu 10-Mar-2016 15:00
* Amsterdam: Thu 10-Mar-2016 21:00
* Moscow: Thu 10-Mar-2016 23:00
* Sydney: Fri 11-Mar-2016 07:00
* Tokyo: Fri 11-Mar-2016 05:00

Or in your local time:
* http://www.timeanddate.com/worldclock/fixedtime.html?msg=Node.js+Foundation+Technical%20Steering%20Committee%20(TSC)+Meeting+2016-03-10&iso=20160310T20
* or http://www.wolframalpha.com/input/?i=8pm+UTC%2C+Mar+10%2C+2016+in+local+time

## Links

* **Minutes Google Doc**: <https://docs.google.com/document/d/1Zt6EnkLghHUxqvn8RKz4ewy-uljIdhqU96KHv_uH6oc>
* _Previous Minutes Google Doc: <https://docs.google.com/document/d/1MW6CXlvxF0IWbb0tg5W70FAdWai5rkQoe_Jt9LhHMqQ>_

## Agenda

Extracted from **tsc-agenda** labelled issues and pull requests from the **nodejs org** prior to the meeting.

### nodejs/TSC

* Flatten project, scope TSC, big changes to structure. [#59](https://github.com/nodejs/TSC/pull/59)
* Add guidance to proposal section [#53](https://github.com/nodejs/TSC/pull/53)
* Preferred meeting tools [#48](https://github.com/nodejs/TSC/issues/48)

## Invited

* @bnoordhuis (TSC)
* @chrisdickinson (TSC)
* @cjihrig (TSC)
* @Fishrock123 (TSC)
* @indutny (TSC)
* @jasnell (TSC)
* @misterdjules (TSC)
* @mikeal (observer)
* @mscdex (TSC)
* @orangemocha (TSC)
* @piscisaureus (TSC)
* @rvagg (TSC)
* @shigeki (TSC)
* @trevnorris (TSC)
* @nebrius (Inclusivity WG Nominee)

## Notes

The agenda comes from issues labelled with `tsc-agenda` across **all of the repositories in the nodejs org**. Please label any additional issues that should be on the agenda before the meeting starts.

## Joining the meeting

[...]

Issue #57, opened 2016-03-03
Node.js Foundation Technical Steering Committee (TSC) Meeting 2016-03-03


rvagg:
## Time

**UTC Thu 03-Mar-2016 20:00**:

* San Francisco: Thu 03-Mar-2016 12:00
* New York: Thu 03-Mar-2016 15:00
* Amsterdam: Thu 03-Mar-2016 21:00
* Moscow: Thu 03-Mar-2016 23:00
* Sydney: Fri 04-Mar-2016 07:00
* Tokyo: Fri 04-Mar-2016 05:00

Or in your local time:
* http://www.timeanddate.com/worldclock/fixedtime.html?msg=Node.js+Foundation+Technical%20Steering%20Committee%20(TSC)+Meeting+2016-03-03&iso=20160303T20
* or http://www.wolframalpha.com/input/?i=8pm+UTC%2C+Mar+03%2C+2016+in+local+time

## Links
* **Minutes Google Doc**: <https://docs.google.com/document/d/1MW6CXlvxF0IWbb0tg5W70FAdWai5rkQoe_Jt9LhHMqQ> (Edit: (Fishrock123): the URLs were wrong so I changed them but now the old link is broken, oopsie) 
* _Previous Minutes Google Doc: <https://docs.google.com/document/d/1S1hzSEB1-_b2cHhcCSXRIiH1irWiWfLS4KawKQMPwvl>_
* **Hangouts On-air** (for participants): https://plus.google.com/hangouts/_/hoaevent/AP36tYe8U3yvYH0PsZbR0qG_3voYfaJ1-G6c1oZqVa9-ftABcBJiQQ
* **YouTube** (live feed for anyone, please be patient, it takes us a while to get started and broadcast): http://www.youtube.com/watch?v=oYrlnRSEF94

## Agenda

Extracted from **tsc-agenda** labelled issues and pull requests from the **nodejs org** prior to the meeting.

### nodejs/TSC

_nil_

We will likely be discussing foundational Foundation things, see some of recent open pull requests on this repo for that theme.

## Invited

* @bnoordhuis (TSC)
* @chrisdickinson (TSC)
* @cjihrig (TSC)
* @Fishrock123 (TSC)
* @indutny (TSC)
* @jasnell (TSC)
* @misterdjules (TSC)
* @mikeal (observer)
* @mscdex (TSC)
* @orangemocha (TSC)
* @piscisaureus (TSC)
* @rvagg (TSC)
* @shigeki (TSC)
[...]

PR #108, opened 2016-06-16
Add @nebrius as a member of the TSC

williamkapke:
I'll start by saying: I probably missed something. 😕 

I know @nebrius was [nominated](https://github.com/nodejs/inclusivity/issues/78) by the @nodejs/inclusivity WG as their rep on the TSC... but he didn't officially make it to the TSC? I can't tell.

He started attending TSC meetings as an observer: #57
Then as the nominee: #61
Since he is just has been listed without `(TSC)` after his name: #69

He is also not part of @nodejs/tsc team.

If this is just needing a PR... well here ya go!


rvagg:
Yes, what you're missing is that I'd expressed a preference to hold off on membership changes to the TSC until we figure out what this group is even for. That process includes trying to make progress on the mission statement, which needs a definition of "node core", then ensuring that the relationship with the board is crystal clear so we know our boundaries, then after that, figuring out what it means to have separate TSC and CTC bodies: what is the nature of the relationship, how do they serve each other, what is there beyond "core" that the TSC might need to care about and how does the TSC fit into that picture in ways that the CTC doesn't? And do we even need a separate TSC or are we just overcomplicating things?

I hope that explains why I've been hesitant to support membership changes, it's nothing about individuals involved by any means, I just don't want to go down a path of entrenching governance bodies that don't have clarity around the reasons for which they exist!

Having said that, @nebrius has been great to have on board and has been very constructive at this level and it's not cool that we started the process of getting an Inclusivity rep on board but never took it to the next step of having a formal vote to make them a TSC member. So let's see if we can get something close to a quorum to have a vote and make it stick. I'm confident that @nebrius is well placed to help us answer some of the questions that I've laid out above.

nebrius:
Thank you for the vote of confidence @rvagg :)

Yes it's been somewhat annoying that the process was started and then stalled, but I completely understand why so don't worry about it. I'm happy we're moving the process forward, and please let me know if there's anything you need from me on my end.

williamkapke:
Thursday's meeting (#110) did not have enough members present for quorum so @nebrius was unable to be added. The 3 members present were in favor. @rvagg said he would try to solicit votes from the other @nodejs/tsc members offline to move this forward.

rvagg:
We have +1's from James, Alexis, Jeremiah and myself.

@bnoordhuis, @chrisdickinson, @cjihrig, @indutny, @misterdjules, @mscdex, @piscisaureus, @shigeki, @trevnorris you're all still on the TSC, we haven't done anything to the membership other than this vote, see above for reasoning on that. In order to move forward here could you all please consider dropping a vote one way or the other in here? This is for the TSC only, doesn't impact the CTC, which already has diverged in its membership.

misterdjules:
LGTM

shigeki:
I'm +1 and sorry for my late vote.

mscdex:
+1

jasnell:
Still LGTM!

PR #154, opened 2016-10-20
Streaming docs

Fishrock123:
finally got this done... shows how to stream to the Node.js youtube with Open Broadcaster Software.

cc @nodejs/tsc @mikeal


williamkapke:
I gave this a cursory review for any obvious feedback- but I don't have any :D 

I think we should pull this in and iterate on it if/when people find any errors. Good work @Fishrock123. 🍻

Fishrock123:
Noted, merging so this can be used for the upcoming CTC meeting. We'll iterate.

Issue #145, opened 2016-09-22
Let's detail the process for how funds are distributed for travel


nebrius:
According to my best understanding, there are three sources of funding available for TSC members
- Community fund, as managed by @feross
- General Foundation funds
- Community Travel Fund (colloquially referred to as the TSC fund)

The process for requesting funds from one of the sources, and which fund should be tapped, is not currently documented clearly. We should create a document that instructs people how to apply for funds and which fund they should apply for.

This issue was motivated by https://github.com/nodejs/TSC/issues/141.

feross:
Community fund is not up and running yet. We still have a bit of work to do to get Node board approval. Hoping to have that all in place by end of year.

yoshuawuyts:
How much funds are available anyhow? We know it should be somewhere upwards of 2 million per year (e.g. gold / silver / bronze sponsor levels, etc.) but numbers don't seem to be publicly available. Perhaps mild side track, but if this is a public chat about money, it'd be cool to know how much there is :tada:

feross:
@yoshuawuyts Yeah, your guess is roughly right. The foundation should really publish this information. cc @mikael

mikeal:
Answers!

> Community fund, as managed by @feross

This is not setup, we still need the final doc worked out and a board resolution passed along with a budget allocation from the general fund.

> Community Travel Fund (colloquially referred to as the TSC fund)

$45K are in this fund, none of it has been touched, and it expires at the end of the year. The TSC can pass resolutions to spend this at its own discretion and any approved travel expenses can be sent to ap@linuxfoundation.org for re-imbursement (CC me as well).

> General Foundation funds

This one is a bit tricky. The foundation was formed mid-year which meant that we had a July - July budget for the first year. We now have a 6 month budget in place that brings us through the end of this year.

Because both of our large events fall in that 6 month period we are over-allocated in the 6 month cycle (as in we will be spending more than we get in revenue). We have enough surplus from the previous cycle to cover it but it's not a great time to do *new* allocations.

In order to get new funds out of the general fund (in 2016 or 2017) it requires a board resolution. Rod, as the TSC chair, can bring a resolution to the board at any time, and I'm on the hook to help him draft it if he needs it.

TheAlphaNerd:
can we use the travel fund to bring ctc members to interactive us? should I
open a separate issue for this?

Fishrock123:
[...]

PR #181, opened 2016-12-25
General update to the Member Travel Fund doc

addaleax:
We talked about travel funding at the last TSC meeting with the result that the document on it should be updated and we would want to keep track of all approved requests (and, in particular, their amounts) in a table.

@nodejs/tsc Please take a look and let me know what you think!

### DIFF
--- a/Member-Travel-Fund.md
+++ b/Member-Travel-Fund.md
@@ -19,6 +19,7 @@ for the TSC to approve or decline.
  * Include the presentation you intend to give, if applicable.
  * Include the size of the stipend you wish to receive.
   * Reimbursement stipends are expected to vary with travel distance.
+* Once the final amount spent is known, update the table again with that information.
 
 The TSC will consider each proposal in its next regular meeting. While it is
 ultimately at the TSC's discretion who to approve the following considerations
--- a/Member-Travel-Fund.md
+++ b/Member-Travel-Fund.md
@@ -1,45 +1,44 @@
 ## Purpose
 
 To establish and administer a fund for members of the Node.js
-Foundation to travel and spread knowledge about the foundation
+Foundation to travel and spread knowledge about and support the foundation
 and the Node.js Foundation projects.
 
 ## Restrictions
 
 * Candidates must be Individual Members of the Node.js Foundation.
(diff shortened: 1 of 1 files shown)


williamkapke:
This was discussed in the 2016-12-15 TSC meeting here:
https://www.youtube.com/watch?v=r1chJO9nbRQ&t=2090s

williamkapke:
This should fix #145

addaleax:
> This should fix #145

Added a `Fixes:` tag to the commit message, thanks for pointing that out!

mikeal:
I don't know if this needs to be in the doc but in the case where we don't provide a stipend but instead do a re-imbursement or book the travel ourselves we should update the doc with the real cost.

addaleax:
@mikeal Sorry for the late reply – I guess that depends on who’s doing the update. I assume you wouldn’t want to take that work on for yourself, but otoh I think in the cases where travel is directly booked by the NF the person in question doesn’t even know the total amount? (I don’t know how much my flight to Austin was, for example.)

addaleax:
I probably won’t be able to be at tomorrow’s TSC meeting, so just to let everyone know:
- I haven’t merged this PR yet because I’d kinda prefer if a TSC member merged this
- I don’t think this is urgent, so feel free to wait on @mikeal’s response to my last comment

Fishrock123:
@addaleax so @mikeal on call today said we can just get that info from the foundation (such as from @mikeal).

Could you update the doc with info stating that one travel is done we should open a PR against the repo to adjust it to the actual cost (including any costs we may need to get from the NF)?

Fishrock123:
@nodejs/tsc Will be merging this once @addaleax updates it if there are no outstanding concerns.

addaleax:
@Fishrock123 done, added a short note on that. if you prefer some other wording let me know, or just edit this PR :)

Fishrock123:
TSC has been consistently positive about this, landing as we all appear to have consensus about this.

Landed in https://github.com/nodejs/TSC/commit/6ea0b676e1e7cc3d9587b9aa3cc26ebf8dcb4f50, thanks again @addaleax!

PR #193, opened 2017-01-11
doc: update the OBS guide doc.

Fishrock123:
Updates to be more consistent, and add the new "Audio Only" text.


Fishrock123:
Landed in https://github.com/nodejs/TSC/commit/4859107e21a5a0885c72f727ab39cc403af07606, not 100% sure why github has weirdness with the merge.

PR #208, opened 2017-02-09
remove myself from the TSC

chrisdickinson:
Hey! I haven't time lately to attend meetings, & it seems unlikely that I'll have time in the near future. In light of that I'm removing myself from the TSC. :wave: So long, and thanks for all the fish! :dolphin:


Fishrock123:
Landed in https://github.com/nodejs/TSC/commit/df45eb0b5204ada6290eea3db17341ddb393ac36, thanks.

PR #225, opened 2017-03-17
docs: add joshgav

joshgav:
Add joshgav to the list of TSC members. Also need to be added to GH groups.


rvagg:
@joshgav pls also review https://github.com/nodejs/email/pull/36 and we'll get you on the email list too

PR #273, opened 2017-05-26
format of moderation policy

Trott:
Indentation of bullets is not rendering properly. Increasing markdown
indentation from one space to two spaces appears to fix this.

## Before
 
screen shot 2017-05-26 at 12 08 15 pm

## After

screen shot 2017-05-26 at 12 08 38 pm


Trott:
(I'm not TSC so I'd prefer if someone on TSC merged this rather than me. I'm guessing it doesn't need to wait 72 hours or anything. It's merely a formatting change.)

nebrius:
yep agreed, we don't need to wait

Issue #261, opened 2017-05-13
Adopting Contributor Covenant as CoC


jasnell:
The [Contributor Covenant](http://contributor-covenant.org/) has emerged as a standard Code of Conduct used by a [large number](http://contributor-covenant.org/adopters/) of open source projects within the larger Node.js ecosystem and other ecosystems. The [CoC](http://contributor-covenant.org/version/1/4/) text itself is both comprehensive and extremely well thought out.

I am proposing that the TSC adopt the Contributor Covenant CoC as the new Code of Conduct for the entire Node.js project, replacing the existing one.

*Update*: @nodejs/collaborators @nodejs/members ... So far there have not yet been any objections raised from the @nodejs/tsc, but given the impact on the entire project, I'd like to solicit input from the broader base of contributors. If you agree with the change you needn't weigh in except maybe with a simple thumbs up emoji. If you object to the change, please add a comment explaining your objection. Please keep comments focused and on topic. Thank you!

ChALkeR:
While using a standard and maintained by someone else version looks like a good idea, there are some details that should be somehow taken care of, and for that we need to understand the actual changes being done by such replacement.

_Probablably_ an incomplete list:

### Pledge

> or other similar characteristic

Is being removed. 

### Contact

Current CoC states:
> Instances of abusive, harassing, or otherwise unacceptable behavior may be
reported by:
> 
> * Emailing [report@nodejs.org](mailto:report@nodejs.org) (this will email all TSC members)
> * Contacting [individual TSC members](https://nodejs.org/en/foundation/tsc/#current-members-of-the-technical-steering-committee).

The Contributor Covenant states:

> Instances of abusive, harassing, or otherwise unacceptable behavior may be
reported by contacting the project team at [INSERT EMAIL ADDRESS].

Either the Contributor Covenant has to be changed in that specific part somehow, or it should be agreed that the second way of geting in touch should be removed. Ref: #45.

### Moderation

Current CoC has a link to the moderation policy.
> See the TSC's [moderation policy](https://github.com/nodejs/TSC/blob/master/Moderation-Policy.md) for details about moderation.

This change would remove that.

### Profanity

https://github.com/nodejs/node/pull/3827 is being reverted. The new CoC does not mention the profanity at all.

mikeal:
@ChALkeR I'd expect the contacts to be updated. Our moderation policy is written as a separate document so I don't think we'll be dropping it with the new CoC.

Good catch on the profanity addition that isn't paired.

[...]

PR #274, opened 2017-05-30
meta: adopt the contributor covenant CoC

jasnell:
Fixes: https://github.com/nodejs/TSC/issues/261

@nodejs/tsc
/cc @nodejs/collaborators

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,56 +1,74 @@
-# Code of Conduct
+# Contributor Covenant Code of Conduct
 
-## Conduct
+## Our Pledge
 
-* We are committed to providing a friendly, safe and welcoming
-  environment for all, regardless of level of experience, gender
-  identity and expression, sexual orientation, disability,
-  personal appearance, body size, race, ethnicity, age, religion,
-  nationality, or other similar characteristic.
-* Please avoid using overtly sexual nicknames or other nicknames that
-  might detract from a friendly, safe and welcoming environment for
-  all.
-* Please be kind and courteous. There's no need to be mean or rude.
-* Respect that some individuals and cultures consider the casual use of
-  profanity and sexualized language offensive and off-putting.
-* Respect that people have differences of opinion and that every
-  design or implementation choice carries a trade-off and numerous
-  costs. There is seldom a right answer.
-* Please keep unstructured critique to a minimum. If you have solid
-  ideas you want to experiment with, make a fork and see how it works.
-* We will exclude you from interaction if you insult, demean or harass
-  anyone. That is not welcome behavior. We interpret the term
-  "harassment" as including the definition in the [Citizen Code of
-  Conduct](http://citizencodeofconduct.org/); if you have any lack of
-  clarity about what might be included in that concept, please read
-  their definition. In particular, we don't tolerate behavior that
-  excludes people in socially marginalized groups.
-* Private harassment is also unacceptable. No matter who you are, if
-  you feel you have been or are being harassed or made uncomfortable
-  by a community member, please contact one of the channel ops or any
-  of the TSC members immediately with a capture (log, photo, email) of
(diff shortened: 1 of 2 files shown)


Moderation-Policy.md
```diff
@@ -43,7 +43,28 @@ Any alternative Moderation Policy used for a given repository must be included i
...
+  professional setting
+
+In addition, the following guidelines apply:
+
+* Please be kind and courteous. There's no need to be mean or rude.
+* Respect that some individuals and cultures consider the casual use of profanity and sexualized language offensive and off-putting.
+* Respect that people have differences of opinion and that every design or implementation choice carries a trade-off and numerous costs. There is seldom a right answer.
+* Please keep unstructured critique to a minimum. If you have solid ideas you want to experiment with, make a fork and see how it works.
+* Insulting, demeaning, or harassing anyone, in public or in private, is not welcome behavior and will be grounds for banning from the project.
+* Use of personal pronouns in code comments and documentation is to be avoided.
```
addaleax: I think the CoC never really was a good place for this statement, and the moderation policy seems even less so… can we move this to doc/STYLE_GUIDE.md in nodejs/node or so?

CODE_OF_CONDUCT.md
```diff
@@ -1,56 +1,74 @@
...
+when an individual is representing the project or its community. Examples of
+representing a project or community include using an official project e-mail
+address, posting via an official social media account, or acting as an appointed
+representative at an online or offline event. Representation of a project may be
+further defined and clarified by project maintainers.
+
+## Enforcement
+
+Instances of abusive, harassing, or otherwise unacceptable behavior may be
+reported by contacting the project team at report@nodejs.org. All
```
addaleax: I agree with the commentary in the issue that it’s not a good thing that readers can’t tell *who* is reachable at this email address (I don’t even know for sure), and that the possibility of reaching out to trusted individuals isn’t mentioned here (which is, at this point, the course of action that I would personally recommend to anyone who wants to report a problem).

CODE_OF_CONDUCT.md
```diff
@@ -1,56 +1,74 @@
...
+complaints will be reviewed and investigated and will result in a response that
+is deemed necessary and appropriate to the circumstances. The project team is
```
thefourtheye: Everywhere else "project maintainers" is used, except here.

jasnell: This text is taken entirely from the Contributor Covenant word for word. Changes would need to be proposed there.

CODE_OF_CONDUCT.md
nebrius: What if we instead used tsc@iojs.org or commcomm@iojs.org (the latter would need to be created)? This would at least indicate which group of people the email would go to, and maybe add a blurb such as `or contact a member of the [TSC](https://github.com/nodejs/TSC#tsc-members) or [CommComm](https://github.com/nodejs/community-committee#community-committee-members) directly`?

jasnell:
I think this is just about ready to merge. The one remaining issue is the email notification. Our current moderation policy points to report@nodejs.org so that's what I think we should go with for now. If we want to change that policy, then ok, but let's get this merged and *then* look at that.

jasnell:
Ping @nodejs/tsc ... one final call for objections. If there are no objections by Friday I will land.

addaleax:
ping @jasnell I don’t know, do you want to address the comments here? If you want to merge this as-is, it should be fine.

jasnell:
Let me take another look this afternoon.

jasnell:
@nodejs/tsc ... final call. I'll land either later today or tomorrow if there are no objections.

jasnell:
@nodejs/collaborators @nodejs/tsc @nodejs/ctc ... given how long this has been open, the signoff and the lack of any objections, I'm calling it approved. Will land this now!

PR #279, opened 2017-06-01
docs: update mentors

MylesBorins:
orangemocha is no longer on TSC, they should no longer be mentioend
as a mentor.

Is the mentor section relevant anymore?


jasnell:
If @orangemocha is not actively involved with libuv now, then this makes sense. I could be wrong but I do not believe there's a requirement that mentors have to be on the TSC.

MylesBorins:
If that is not a requirement then feel free to close 😄

Issue #312, opened 2017-08-22
Re-merge TSC and CTC


jasnell:
@nodejs/ctc @nodejs/tsc - We have discussed this for quite some time now. There is little continued reason for the TSC and CTC to continue as separate entities. I propose that we merge the CTC and TSC back into a single body. Essentially, what this means is (a) dechartering the CTC and (b) adding all current members of the CTC to the TSC. No TSC Charter changes are required. Some documentation would need to be updated, and the existing nodejs/ctc repository would be archived.

cjihrig:
+1

jasnell:
Here's my +1

MylesBorins:
+!

Fishrock123:
+1

mhdawson:
+1

indutny:
+1

jasnell:
@nodejs/ctc and @nodejs/tsc members, while leaving emoji responses is fine in general, for this issue please indicate your position explicitly in a comment. I will prepare the necessary pull requests to kick this process off tomorrow.

mscdex:
+1

ofrobots:
+1.

joyeecheung:
+1

ChALkeR:
+1. I think that I liked the earlier proposal at https://github.com/nodejs/TSC/pull/263 more, though, but the change discussed here looks better than the current situation.

flickz:
+1

mcollina:
I'm +1.

shigeki:
+1

Trott:
I'm not opposed to this (yet), but I do think some additional things need to happen at the same time or else this is going to cause more problems than it will solve.

I'm going to greatly simplify things here (e.g., splitting problems into "technical" problems vs. "human" problems).

* As it stands now, this will basically result in the TSC being CTC + Josh Gavant. That's more representative of the project as a whole (in my opinion anyway) than the current remaining 9 members of the TSC, but...

[...]

Issue #311, opened 2017-08-21
Document guidelines for accountability expectations for all members of the TSC, CTC, and CommComm o


mhdawson:
This issue is to discuss/document the expectations that we have on our TSC, CTC and ComCom members in addition to those outline in the existing
[code of conduct](https://github.com/nodejs/TSC/blob/master/CODE_OF_CONDUCT.md).  

It is important that members of these groups act in a way that not only complies with the code of conduct but in a way that both demonstrates support and strives to avoid potential impression that those helping to lead the efforts of these groups and Node.js community feel otherwise.

It is important that we document these expectations so that they can be communicated to members as they join these groups. 

I received a good suggestion that the board may have a doc that could provide a starting point.  I looked and found https://github.com/nodejs/board/pull/50 but I'll keep looking in case there is something more.

So which might apply include (adjusted of course to apply to TSC, CTC and ComCom):

 +## Support of Board Decisions
 +
 +A Director must accept and publicly support Board decisions. A Director is encouraged to be an ambassador of the Foundation and, subject to the Director’s confidentiality and other obligations under this Code of Conduct, to promote the activities and actions of the Board with the Foundation membership and publicly. In doing so, a Director must stay faithful to the intent of the Board as expressed in its official statements, and should not reinterpret or re-characterize the Board’s actions to reflect the Director’s own view. 
 +
 +While a Director has the right and responsibility to exercise independent judgment and to express dissenting opinions during Board deliberations, a Director also has the obligation outside the Boardroom to respect and support decisions of the majority, even when the Director dissented from the majority view. A Director who does not support a Board decision may express the Director’s opposition within the Board in an appropriate manner, but must not take actions publicly or with respect to the Foundation membership that have the purpose or result of undermining the decisions or actions of the Board. Acting otherwise may be in violation of a Director’s fiduciary and loyalty duties. Accordingly, a Director who intends to publicly oppose a Board action should resign the Director’s position on the Board before doing so.
 +

Over the next few days I'm hoping we can work through some ideas and put together a draft of the proposed text.

@nodejs/collaborators

stephenburgess8:
>It is important that we document these expectations so that they can be communicated to members as they join these groups.

[I wrote up my idea for a proposal](https://github.com/nodejs/TSC/pull/318#issuecomment-324213444) that would provide a documented mechanic or standard by which issues relating to members breaking policy could be handled. In short, TSC members could issue censures and limit the responsibilities of other TSC members.

[...]

PR #318, opened 2017-08-22
charter: add TSC membership terms, mandatory mediation, and removal requirements

jasnell:
This PR proposes a change to the TSC Charter. It would need to be approved by the TSC and the Node.js Foundation Board.

Specifically, this introduces a couple of changes:

* 1 year terms for TSC members that must be recertified each year by a simple majority of the TSC peers.

* Adds a requirement that TSC members must maintain participation in at least one chartered working group or team through their term

* Adds mandatory independent, binding third-party mediation under the oversight of the Node.js Foundation Board should any TSC member be reported for a Code of Conduct or Policy violation.

@nodejs/tsc @nodejs/ctc @hackygolucky

### DIFF
--- a/TSC-Charter.md
+++ b/TSC-Charter.md
@@ -220,12 +220,6 @@ overturn or vacate the decision of the TSC, the Board may issue a statement
 requesting that the TSC revisit the decision and recommend that the matter be
 referred to a Foundation appointed third party mediator for arbitration.
 
-All decisions regarding removal of TSC members is subject to review by the body
-of Node.js collaborators. Should no fewer than one-quarter of the current
-Node.js project Collaborators (as defined by the TSC's governance and
-contribution policies) disagree with the TSC vote, the matter shall be referred
-to a Foundation appointed third party mediator for arbitration. 
-
 Once referred to arbitration, the decision of the mediator will be considered
 final and binding on all parties.
 
--- a/TSC-Charter.md
+++ b/TSC-Charter.md
@@ -36,10 +36,12 @@ approved by the Board.
 
 ## Section 4. Establishment of the TSC.
 
-TSC memberships are for one year terms that must be recertified by a simple
-majority of TSC members every year. Recertified members will continue through
-the next years term. Members failing recertification may return to the TSC
-after no less than six months following normal TSC motion.
+TSC memberships are for one year terms that must be recertified. Motions for
+recertification are automatic and require a vote. Members are recertified if a
+simple majority of TSC members vote in favor of recertifying the individual as
+a TSC member. Such members will continue through the next years term.
+Individuals failing recertification may return to the TSC after no less than
+six months following normal TSC motion.
 
 There is no maximum size of the TSC. The size is expected to vary in order to
 ensure adequate coverage of important areas of expertise, balanced with the
@@ -184,6 +186,11 @@ discussion will continue.
 
 For all votes, a simple majority of all TSC members for, or against, the issue
 wins. A TSC member may choose to participate in any vote through abstention.
+
+While the results of all votes must be made public, the actual individual
+ballots cast for most votes may be made public or confidential at the discretion
+of the TSC Chair. However, individual ballots for recertification or removal of
+TSC members must remain confidential.
 
 Note that, in addition to requiring a simple majority vote of the TSC, all
 changes to this charter are also subject to approval from the Node.js
--- a/TSC-Charter.md
+++ b/TSC-Charter.md
@@ -36,10 +36,10 @@ approved by the Board.
 
 ## Section 4. Establishment of the TSC.
 
-TSC memberships are limited to 2 consecutive year periods. An individual
-members term starts at 12am GMT the day after their nomination to the TSC
-is approved and ends precisely two years later. Individuals may be nominated
-to the TSC multiple times but at least one year must pass between terms.
+TSC memberships are for one year terms that must be recertified by a simple
+majority of TSC members every year. Recertified members will continue through
+the next years term. Members failing recertification may return to the TSC
+after no less than six months following normal TSC motion.
 
 There is no maximum size of the TSC. The size is expected to vary in order to
 ensure adequate coverage of important areas of expertise, balanced with the
--- a/TSC-Charter.md
+++ b/TSC-Charter.md
@@ -191,7 +191,8 @@ While the results of all votes must be made public, the actual individual
 ballots cast for most votes may be made public or confidential at the discretion
 of the TSC Chair. However, individual ballots for recertification or removal of
 TSC members must remain confidential. The names of members voting or abstaining
-in all votes must be made public.
+in all votes must be made public with a clear indication given of whether the
+individual voted or abstained.
 
 Following the completion of all votes, a public statement must be made via
 GitHub specifying the results of the vote. Contextual detail about why the
@@ -227,30 +228,28 @@ shall participate in the vote and the remaining members of the TSC may meet or
 discuss the issue privately without the member in question present.
 
 Should a vote to remove a member result in a tie (50%/50% split of non
-abstaining participants) the matter shall be referred to a Node.js Foundation
+abstaining participants) the matter shall be referred to a Foundation
 appointed independent third party mediator for arbitration.
 
 At any point during this process, the individual member in question, the
 individuals opening the request, or the TSC Chair may request that the issue
-be referred to a Node.js Foundation appointed third party mediator for
+be referred to a Foundation appointed third party mediator for
 arbitration.
 
 All decisions regarding removal of TSC members is subject to review by the
-Node.js Foundation Board. Should the Board decide to vacate the TSC vote, the
-matter shall be referred to a Node.js Foundation appointed third party mediator
-for arbitration.
+Node.js Foundation Board. While the Board will not have the ability to
+overturn or vacate the decision of the TSC, the Board may issue a statement
+requesting that the TSC revisit the decision and recommend that the matter be
+referred to a Foundation appointed third party mediator for arbitration.
 
 All decisions regarding removal of TSC members is subject to review by the body
 of Node.js collaborators. Should no fewer than one-quarter of the current
 Node.js project Collaborators (as defined by the TSC's governance and
 contribution policies) disagree with the TSC vote, the matter shall be referred
-to a Node.js Foundation appointed third party mediator for arbitration. 
+to a Foundation appointed third party mediator for arbitration. 
 
 Once referred to arbitration, the decision of the mediator will be considered
 final and binding on all parties.
-
-The third party mediator selected must not be a member of either the TSC,
-Node.js Community Committee, or Node.js Foundation Board of Directors.
 
 ## Section 9. Project Roles
 
@@ -295,12 +294,18 @@ that is established by the TSC. The TSC is required to establish a policy
 for the enforcement of Code of Conduct issues. Any reports of Code of Conduct
 or policy violations on the part of TSC members that are not resolvable through
 that established policy will be referred to binding, independent third party
-mediation under the oversight of the Node.js Foundation Board. All TSC members,
-upon acceptance of their nomination to the TSC body, voluntarily agree to abide
-by the decisions of the independent third party mediator.
+arbitration with a Foundation appointed mediator All TSC members, upon
+acceptance of their nomination to the TSC body, voluntarily agree to abide by
+the decisions of the independent third party mediator.
 
-The third party mediator selected must not be a member of either the TSC,
-Node.js Community Committee, or Node.js Foundation Board of Directors.
+## Section 12: Selection of Mediators
+
+When it becomes necessary within the context of this charter to select a
+third party mediator for arbitration, the mediator shall be selected by the
+Node.js Foundation Board of Directors (or an individual designated by the Board
+to so act on their behalf). The only requirement for this mediator is that the
+individual selected must not be a member of the TSC, Node.js Community
+Committee, or Node.js Foundation Board of Directors.
 
 [Consensus Seeking]: http://en.wikipedia.org/wiki/Consensus-seeking_decision-making
 [Condorcet]: http://en.wikipedia.org/wiki/Condorcet_method
--- a/TSC-Charter.md
+++ b/TSC-Charter.md
@@ -39,7 +39,7 @@ approved by the Board.
 TSC memberships are for one year terms that must be recertified. Motions for
 recertification are automatic and require a vote. Members are recertified if a
 simple majority of TSC members vote in favor of recertifying the individual as
-a TSC member. Such members will continue through the next years term.
+a TSC member. Such members will continue through the next year's term.
 Individuals failing recertification may return to the TSC after no less than
 six months following normal TSC motion.
 
--- a/TSC-Charter.md
+++ b/TSC-Charter.md
@@ -205,20 +205,21 @@ Foundation board.
 
 ### Votes to Remove Members
 
-Should a vote to remove a member result in a tie (50%/50% split of non
-abstaining participants) the matter shall be referred to a Foundation
-appointed independent third party mediator for arbitration.
+Should a situation arise that considers the removal of a TSC member, an
+attempt should first be made to resolve the issue without requiring a vote.
+The TSC Chair may, at their discretion, request that the issue be referred
+to a Foundation appointed third-party mediator for arbitration.
 
-At any point prior to calling for a vote, the individual member in question, the
-individuals opening the request, or the TSC Chair may request that the issue
-be referred to a Foundation appointed third party mediator for
-arbitration.
+Failing all other possible resolutions, a simple majority vote of TSC members
+is required to remove a member. Should such a vote to result in a tie (50%/50%
+split of non-abstaining participants) the matter shall be referred to a
+Foundation appointed independent third-party mediator for arbitration.
 
-All decisions regarding removal of TSC members is subject to review by the
-Node.js Foundation Board. While the Board will not have the ability to
-overturn or vacate the decision of the TSC, the Board may issue a statement
+All decisions regarding removal of TSC members will be subject to review by
+the Node.js Foundation Board. While the Board will not have the ability to
+overturn or vacate the decisions of the TSC, the Board may issue a statement
 requesting that the TSC revisit the decision and recommend that the matter be
-referred to a Foundation appointed third party mediator for arbitration.
+referred to a Foundation appointed third-party mediator for arbitration.
 
 Once referred to arbitration, the decision of the mediator will be considered
 final and binding on all parties.
@@ -265,12 +266,25 @@ Participation in the Node.js project is governed by a Code of Conduct policy
 that is established by the TSC. The TSC is required to establish a policy
 for the enforcement of Code of Conduct issues. Any reports of Code of Conduct
 or policy violations on the part of TSC members that are not resolvable through
-that established policy will be referred to binding, independent third party
+that established policy will be referred to binding, independent third-party
 arbitration with a Foundation appointed mediator. All TSC members, upon
 acceptance of their nomination to the TSC body, voluntarily agree to abide by
-the decisions of the independent third party mediator.
+the decisions of the independent third-party mediator.
 
-## Section 12: Selection of Mediators
+## Section 12: Leaves of Absence and Temporary Suspension of TSC Membership
+
+TSC Members may voluntarily temporarily suspend their memberships in the TSC
+for any period of time not in excess of twelve months. In such cases, the member
+may return to full TSC membership status at any time.
+
+TSC Members may have their memberships involuntarily suspended by a simple
+majority vote of the other TSC members for any period of time not in excess of
+twelve months. The suspension period must be specified in advance of the
+decision. The members privileges as a TSC member will be restored automatically
+after the given period has passed. The TSC may decide, through simple majority
+vote, to lift the suspension early.
+
+## Section 13: Selection of Mediators
 
 When it becomes necessary within the context of this charter to select a
 third party mediator for arbitration, a qualified mediator shall be selected by
--- a/TSC-Charter.md
+++ b/TSC-Charter.md
@@ -228,11 +228,13 @@ represent the Core Project on the TSC.
 ## Section 11. Escalation of Disputes and Code of Conduct Violations
 
 Participation in the Node.js project is governed by a Code of Conduct policy
-that is established by the TSC. Any accusations of Code of Conduct or policy
-violations on the part of TSC members will be referred to binding, independent
-third party mediation under the oversight of the Node.js Foundation Board. All
-TSC members, upon acceptance of their nomination to the TSC body, voluntarily
-agree to abide by the decisions of the independent third party mediator.
+that is established by the TSC. The TSC is required to established a policy
+for the enforcement of Code of Conduct issues. Any reports of Code of Conduct
+or policy violations on the part of TSC members that are not resolvable through
+that established policy will be referred to binding, independent third party
+mediation under the oversight of the Node.js Foundation Board. All TSC members,
+upon acceptance of their nomination to the TSC body, voluntarily agree to abide
+by the decisions of the independent third party mediator.
 
 [Consensus Seeking]: http://en.wikipedia.org/wiki/Consensus-seeking_decision-making
 [Condorcet]: http://en.wikipedia.org/wiki/Condorcet_method
--- a/TSC-Charter.md
+++ b/TSC-Charter.md
@@ -205,28 +205,6 @@ Foundation board.
 
 ### Votes to Remove Members
 
-Requests to remove a member from the TSC may be raised only by other members of
-the TSC or by members of the Node.js Foundation Board. When an issue to remove
-someone from the TSC is raised, the issue must include a list of the specific
-issues the TSC is being asked to consider as grounds for removal. The discussion
-and vote will be limited strictly to considering of the specific issues listed
-in the issue. The issue must also include a description of what outcome those
-reporting the issue expect.
-
-The vote must be structured in terms of whether or not the specific issues
-listed constitute grounds for removal.
-
-When votes to remove individuals from the TSC are scheduled, the member in
-question is to be notified about the vote in advance along with adequate
-contextual details of why the vote is taking place, what, if any, specific
-complaints are being leveled, and the specific questions that the members are
-being asked to consider. The member in question shall be given no less than one
-week before the vote to provide an answer or rebuttal to those issues.
-
-Neither the member in question, nor the member or members raising the issue
-shall participate in the vote and the remaining members of the TSC may meet or
-discuss the issue privately without the member in question present.
-
 Should a vote to remove a member result in a tie (50%/50% split of non
 abstaining participants) the matter shall be referred to a Foundation
 appointed independent third party mediator for arbitration.
@@ -301,11 +279,11 @@ the decisions of the independent third party mediator.
 ## Section 12: Selection of Mediators
 
 When it becomes necessary within the context of this charter to select a
-third party mediator for arbitration, the mediator shall be selected by the
-Node.js Foundation Board of Directors (or an individual designated by the Board
-to so act on their behalf). The only requirement for this mediator is that the
-individual selected must not be a member of the TSC, Node.js Community
-Committee, or Node.js Foundation Board of Directors.
+third party mediator for arbitration, a qualified mediator shall be selected by
+the Node.js Foundation Board of Directors (or an individual designated by the
+Board to so act on their behalf). The mediator should be trained and qualified
+in mediation/conflict resolution and must not be a member of the TSC, Node.js
+Community Committee, or Node.js Foundation Board of Directors.
 
 [Consensus Seeking]: http://en.wikipedia.org/wiki/Consensus-seeking_decision-making
 [Condorcet]: http://en.wikipedia.org/wiki/Condorcet_method
--- a/TSC-Charter.md
+++ b/TSC-Charter.md
@@ -190,11 +190,67 @@ wins. A TSC member may choose to participate in any vote through abstention.
 While the results of all votes must be made public, the actual individual
 ballots cast for most votes may be made public or confidential at the discretion
 of the TSC Chair. However, individual ballots for recertification or removal of
-TSC members must remain confidential.
+TSC members must remain confidential. The names of members voting or abstaining
+in all votes must be made public.
+
+Following the completion of all votes, a public statement must be made via
+GitHub specifying the results of the vote. Contextual detail about why the
+vote was held, including a listing of the specific questions voted on must be
+included in the statement.
 
 Note that, in addition to requiring a simple majority vote of the TSC, all
 changes to this charter are also subject to approval from the Node.js
 Foundation board.
+
+### Votes to Remove Members
+
+Requests to remove a member from the TSC may be raised only by other members of
+the TSC or by members of the Node.js Foundation Board. When an issue to remove
+someone from the TSC is raised, the issue must include a list of the specific
+issues the TSC is being asked to consider as grounds for removal. The discussion
+and vote will be limited strictly to considering of the specific issues listed
+in the issue. The issue must also include a description of what outcome those
+reporting the issue expect.
+
+The vote must be structured in terms of whether or not the specific issues
+listed constitute grounds for removal.
+
+When votes to remove individuals from the TSC are scheduled, the member in
+question is to be notified about the vote in advance along with adequate
+contextual details of why the vote is taking place, what, if any, specific
+complaints are being leveled, and the specific questions that the members are
+being asked to consider. The member in question shall be given no less than one
+week before the vote to provide an answer or rebuttal to those issues.
+
+Neither the member in question, nor the member or members raising the issue
+shall participate in the vote and the remaining members of the TSC may meet or
+discuss the issue privately without the member in question present.
+
+Should a vote to remove a member result in a tie (50%/50% split of non
+abstaining participants) the matter shall be referred to a Node.js Foundation
+appointed independent third party mediator for arbitration.
+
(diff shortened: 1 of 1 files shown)


Trott:
I don't think term limits are necessary, although I won't object if everyone else thinks they're a good idea.

I *do* however think that *terms* are a good idea regardless of whether we impose term limits. You should need to be re-approved by the rest of the TSC every year or so. It's good to remind people to consider whether they want to continue on the committee every so often, rather than leaving it up to them to figure out when to consider if they want to resign their appointment-for-life. As we've seen in practice, people sometimes wait a loooong time after becoming inactive before they decide to step down.

jasnell:
I'm good either way. I prefer the term limits as it sends a stronger message that this is not a permanent standing body but a rotating one that should evolve as the project evolves. It also enforces a period of down time and *helps* reduce the pressure that exists from employers that their people remain in the role -- that is, the term is two years, regardless of what your employer may want, etc.

mcollina:
I agree with @Trott.

I'm 👍  on the other two steps. 

@jasnell the situation is a bit convoluted atm and it's not clear if @nodejs/ctc have a vote on this. Can we wait for #312?

ChALkeR:
@jasnell 

I'm not sure about the term limits for the CTC tasks (and this would have an effect on CTC once CTC and TSC merge).

1) There are different areas of expertise for the issues discussed in CTC, and I think that negative effect, e.g. from removing members of the @nodejs/crypto team (which is small) from the merged TSC/CTC would be more considerable than the benefits here.

2) Unless a better procedure is defined, accepting this in the current form would mean removing most of the people from the merged CTC/TSC the same moment, for a year, if I understand things correctly.

jasnell:
As I've said, I've no problem with relaxing the term limits. However, I want to find an approach that moves us away from this idea that TSC/CTC membership is a permanent status. I'll work up some wording.

jasnell:
Please take a look

TSC-Charter.md
```diff
@@ -216,6 +225,15 @@ TSC.
 * **Maintainer**: a Collaborator within a Core Project elected to
 represent the Core Project on the TSC.
 
+## Section 11. Escalation of Disputes and Code of Conduct Violations
+
+Participation in the Node.js project is governed by a Code of Conduct policy
+that is established by the TSC. Any accusations of Code of Conduct or policy
```
MylesBorins: This seems like it could be subject to abuse if "any" accusation leads to third party mediation.

Perhaps `Code of Conduct violations that involve TSC members that cannot be solved via consensus will be referred to binding, independent third party mediation under the oversight of the Node.js Foundation Board.

ChALkeR:
The new wording regarding the terms looks good to me.

TSC-Charter.md
jasnell: We end up in the same boat with the same hand wringing, delays and interpersonal issues. Going immediately to mediation is better. It doesn't have to be involved. This essentially just means that the Foundation picks someone to look at the issue and decide.

TSC-Charter.md
```diff
@@ -36,10 +36,10 @@ approved by the Board.
 
 ## Section 4. Establishment of the TSC.
 
-TSC memberships are limited to 2 consecutive year periods. An individual
-members term starts at 12am GMT the day after their nomination to the TSC
-is approved and ends precisely two years later. Individuals may be nominated
-to the TSC multiple times but at least one year must pass between terms.
+TSC memberships are for one year terms that must be recertified by a simple
+majority of TSC members every year. Recertified members will continue through
+the next years term. Members failing recertification may return to the TSC
```
Trott: Nit: `years` -> `year's`

Question: What does "failing recertification" mean? Does that mean they forget to put in a motion to get recertified? Or does that mean the TSC opted not to renew their term? Whatever the answer, can we clarify the wording to that effect?

TSC-Charter.md
Trott: Nit: Instead of `accusations of Code of Conduct or policy violations`, it may be less loaded to describe them as `reports of Code of Conduct or policy violations` or `complaints` rather than `accusations`.

TSC-Charter.md
jasnell: What do you recommend here?

TSC-Charter.md
jasnell: I've updated the language here.

TSC-Charter.md
```diff
@@ -216,6 +225,17 @@ TSC.
 * **Maintainer**: a Collaborator within a Core Project elected to
 represent the Core Project on the TSC.
 
+## Section 11. Escalation of Disputes and Code of Conduct Violations
+
+Participation in the Node.js project is governed by a Code of Conduct policy
+that is established by the TSC. The TSC is required to established a policy
```
evanlucas: s/established/establish/

TSC-Charter.md
jasnell: @Trott, I've updated the language. Please let me know if I've Whacked-The-Mole adequately.

Trott: Yes, thanks.

Fishrock123:
I would prefer if a separate appointed team not of the TSC/CTC/CommComm was in change of regular moderation and the "mandatory mediation" be reserved for conflicts internal to that team or _perhaps_ with the committees (but I worry that would disempower the said moderation team).

Fishrock123:
That being said, the "mandatory mediation" is certainly better than any process about that as it is today.

jasnell:
All.. I have updated the proposal to include specific requirements for removal votes that I feel are critical moving forward.

jasnell:
@Fishrock123 ... the updated Moderation policy proposal establishes a separate Moderation Team whose members are subject to approval by both the TSC and CommComm. The members must be Collaborators but are not limited to TSC/CommComm members. We should not have a requirement that Moderation Team members are not TSC/CommComm members because that limits our ability to nominate new members from that team to the TSC or CommComm, and I do not believe that is something we should do.

The Moderation Team needs to be empowered to moderate both the TSC and CommComm. The policy, however, would state that disputes involving either would automatically go to arbitration, providing that higher bar that I believe you're looking for.

TSC-Charter.md
```diff
@@ -176,10 +187,71 @@ discussion will continue.
...
+listed constitute grounds for removal.
+
+When votes to remove individuals from the TSC are scheduled, the member in
+question is to be notified about the vote in advance along with adequate
+contextual details of why the vote is taking place, what, if any, specific
+complaints are being leveled, and the specific questions that the members are
+being asked to consider. The member in question shall be given no less than one
+week before the vote to provide an answer or rebuttal to those issues.
+
+Neither the member in question, nor the member or members raising the issue
```
evanlucas: I'm not sure I agree with this...it seems like that increases the likelihood of the member in question *not* being removed. Especially if more than one member raises the issue. 

Example:

Say there are 10 members on the TSC. 4 members step forward and raise an issue to remove another member. This would mean that there are only 5 votes that can be cast. If 3 members vote against, and 2 members vote for, the vote would not pass, which to me seems....quite problematic?

gibfahn:
>The members must be Collaborators but are not limited to TSC/CommComm members.

So [Core collaborators](https://github.com/orgs/nodejs/teams/collaborators/members) not [org members](https://github.com/orgs/nodejs/teams/members/members) right?

jasnell:
Yes

TSC-Charter.md
jasnell: Potentially, but it also justifies referring the issue to arbitration, which can be requested either by those submitting the request, the person being considered for removal, or the TSC Chair. The decision can also be vacated by the Collaborator base or the Node.js Foundation Board, so there are contingencies and mitigations included.

The purpose of this clause is to avoid the possibility of railroading an individual off the committee and I believe it's necessary.

TSC-Charter.md
```diff
@@ -36,16 +36,27 @@ approved by the Board.
...
+TSC memberships are for one year terms that must be recertified. Motions for
+recertification are automatic and require a vote. Members are recertified if a
+simple majority of TSC members vote in favor of recertifying the individual as
+a TSC member. Such members will continue through the next year's term.
+Individuals failing recertification may return to the TSC after no less than
+six months following normal TSC motion.
+
+There is no maximum size of the TSC. The size is expected to vary in order to
+ensure adequate coverage of important areas of expertise, balanced with the
+ability to make decisions efficiently. The TSC must have at least three members.
```
gr2m: I would suggest a higher number, at least 5, given the TSC’s responsibility 3 people is to few. I would also suggest to document what happens if the number falls below the minimum.

Another topic that might fit better into a follow up pull request is to require "active" members, and define what "active" means, to avoid a situation where people are voted into the board without fulfilling their responsibilities.

jasnell: -1 on making the minimum five. What happens if the number falls below this is the responsibility of the Node.js Foundation Board to determine.

jasnell: Regarding "active" members, the TSC recently passed modification to the TSC Charter that includes specific guidelines around TSC member activity. Specifically, there is a minimum participation requirement already covered by the charter.

Trott: > Another topic that might fit better into a follow up pull request is to require "active" members, and define what "active" means, to avoid a situation where people are voted into the board without fulfilling their responsibilities.

@gr2m FYI, [this section of the charter](https://github.com/nodejs/TSC/blob/master/TSC-Charter.md#section-4-establishment-of-the-tsc) includes these minimum requirements to stay active:

> In the case where an individual TSC member -- within any three month period -- attends fewer than 25% of the regularly scheduled meetings, does not participate in TSC discussions, and does not participate in TSC votes, the member shall be automatically removed from the TSC. The member may be invited to continue attending TSC meetings as an observer.

(Not saying there isn't room for improvement. Just pointing out what's already there.)

TSC-Charter.md
mhdawson: Would an option be to allow both the raisers and the member in question.  If its one raiser then they will balance out, otherwise more raisers end up with more weight ?

TSC-Charter.md
```diff
@@ -176,10 +187,71 @@ discussion will continue.
 For all votes, a simple majority of all TSC members for, or against, the issue
 wins. A TSC member may choose to participate in any vote through abstention.
 
+While the results of all votes must be made public, the actual individual
+ballots cast for most votes may be made public or confidential at the discretion
+of the TSC Chair. However, individual ballots for recertification or removal of
+TSC members must remain confidential. The names of members voting or abstaining
+in all votes must be made public.
```
mhdawson: I assume it would not be identified if they voted or abstained, correct ?

jasnell: No, my assumption is that there would be clear indication who voted and who abstained.

TSC-Charter.md
```diff
@@ -36,16 +36,27 @@ approved by the Board.
...
+Individuals failing recertification may return to the TSC after no less than
+six months following normal TSC motion.
+
+There is no maximum size of the TSC. The size is expected to vary in order to
+ensure adequate coverage of important areas of expertise, balanced with the
+ability to make decisions efficiently. The TSC must have at least three members.
+
+Individuals nominated to the TSC must be members of at least one Node.js
+Working Group at the time of their nomination and must maintain active
+participation in at least one Working Group throughout their term.
```
mhdawson: This one concerns me a bit.  For some areas we have teams as opposed to WGs (ex crypto) and I hate to lose those voices from the TSC.  We should look at how many of the current TSC members would be affected by this requirement as I think it would b quite a high number.

mhdawson:
Overall in favor with a few comments as noted.

TSC-Charter.md
jasnell: The goal here is to strongly encourage TSC members to be more active across the project. Those individuals could easily jump into contributing to Working Groups also

gibfahn: I guess this also means that people who only contribute to core are excluded.

Maybe it could say that they must participate in the Node.js org throughout their term.

mcollina: How about `Working Group or Team`?

jasnell: I'd be good with that.

TSC-Charter.md
```diff
@@ -176,10 +187,71 @@ discussion will continue.
...
+complaints are being leveled, and the specific questions that the members are
+being asked to consider. The member in question shall be given no less than one
+week before the vote to provide an answer or rebuttal to those issues.
+
+Neither the member in question, nor the member or members raising the issue
+shall participate in the vote and the remaining members of the TSC may meet or
+discuss the issue privately without the member in question present.
+
+Should a vote to remove a member result in a tie (50%/50% split of non
+abstaining participants) the matter shall be referred to a Node.js Foundation
```
mcollina: Is this going to the board? Then it needs to be said explicitly. Who appoints it from the Node.js Foundation?

jasnell: The Node.js Board or the Executive Director on behalf of the Board.

TSC-Charter.md
```diff
@@ -176,10 +187,71 @@ discussion will continue.
...
+of Node.js collaborators. Should no fewer than one-quarter of the current
+Node.js project Collaborators (as defined by the TSC's governance and
+contribution policies) disagree with the TSC vote, the matter shall be referred
+to a Node.js Foundation appointed third party mediator for arbitration. 
+
+Once referred to arbitration, the decision of the mediator will be considered
+final and binding on all parties.
+
+The third party mediator selected must not be a member of either the TSC,
+Node.js Community Committee, or Node.js Foundation Board of Directors.
```
mcollina: So it can be one of the collaborators that disagree with the TSC vote? IMHO at this point they should be external to the collaborators as well.

TSC-Charter.md
mcollina: Can you write down a paragraph on how the mediator is nominated, so that if follows the same pattern everywhere?

TSC-Charter.md
jasnell: The Board (or Executive Director on the Boards behalf) would select the individual. This is just an additional constraint.

TSC-Charter.md
jasnell: Will do

TSC-Charter.md
jasnell: No, I think that ends up creating an imbalance and does not avoid the possibility of railroading the decision.

jasnell:
Updated. PTAL

TSC-Charter.md
```diff
@@ -176,10 +187,71 @@ discussion will continue.
...
+Following the completion of all votes, a public statement must be made via
+GitHub specifying the results of the vote. Contextual detail about why the
+vote was held, including a listing of the specific questions voted on must be
+included in the statement.
+
 Note that, in addition to requiring a simple majority vote of the TSC, all
 changes to this charter are also subject to approval from the Node.js
 Foundation board.
 
+### Votes to Remove Members
```
Trott: I know you worked hard on this section and it's a lot of text and I'm sure it represents a lot of thought. But I disagree with it and think it should simply be removed. Go back to the original text on the matter which was something like:

> TSC members may be removed by a standard TSC motion.

Trying to figure out a process ahead of time is bound to fail, especially if it's people like us trying to figure it out.

I get that this is trying to address recent events, but IMO the issue there isn't that the TSC didn't know how to remove someone. The how-to-remove-someone process is actually straightforward. It's that the TSC didn't know how to deal with CoC stuff.

If we could do it all over again, I might suggest that the whole thing be handled like a standard TSC motion. If someone wants another member removed, make a motion. A discussion ensues. If consensus cannot be reached, it goes to a vote. There does not need to be misconduct for someone to be removed. Someone can be removed for any number of reasons. Maybe they are inactive. Maybe they are difficult to work with, without being malicious. Maybe they consistently cause problems unintentionally. Maybe they just have bad judgment. Maybe it's a combination of things like that, any one of which wouldn't be a big deal, but taken together make them unsuitable for the TSC.

Creating artificial barriers for removal ("vote must be structured in terms of whether or not the specific issues listed constitute grounds for removal") seems counterproductive. If anything, it has been too *hard* to remove members from the TSC or CTC. As evidence, it has never happened, even in one case when someone was basically inactive for 2 years.

If we want to include additional process to try to prevent a repeat of this experience, maybe focus on referring disputes to third-party mediation (by someone trained in mediation/conflict resolution, not some other member of this or another committee or even the board).  And ask the Board to agree in advance that they're willing to pay for such things.

TSC-Charter.md
```diff
@@ -176,10 +187,71 @@ discussion will continue.
...
+arbitration.
+
+All decisions regarding removal of TSC members is subject to review by the
+Node.js Foundation Board. Should the Board decide to vacate the TSC vote, the
+matter shall be referred to a Node.js Foundation appointed third party mediator
+for arbitration.
+
+All decisions regarding removal of TSC members is subject to review by the body
+of Node.js collaborators. Should no fewer than one-quarter of the current
+Node.js project Collaborators (as defined by the TSC's governance and
```
Trott: Is this referring to Collaborators as we currently use the term or the redefined Collaborators to mean everyone in the GitHub org? The latter could be really problematic because I'd guess that between 40% and 60% of those folks aren't actually active on the project in any significant way.

TSC-Charter.md
evanlucas: I believe that the way this is currently worded creates an imbalance. I'm not seeing this is a railroading of the decision either. If 4 people come forward with issues with another member, then those four people don't get to vote on whether or not to remove the individual. For any other matter, one can simply call a vote and their votes be counted. Why does it have to be different here? This would make it even more difficult to remove a toxic member from the project. To me, this *enables* one of the problems it is trying to solve even more.

jasnell: Ok, noted. I'll pull this back out.

TSC-Charter.md
jasnell: > I know you worked hard on this ...

That's not even a consideration for me.

I will rework this.

jasnell:
@Trott and @evanlucas ... please take a look now.

TSC-Charter.md
```diff
@@ -176,10 +187,48 @@ discussion will continue.
...
+All decisions regarding removal of TSC members is subject to review by the
+Node.js Foundation Board. While the Board will not have the ability to
+overturn or vacate the decision of the TSC, the Board may issue a statement
+requesting that the TSC revisit the decision and recommend that the matter be
+referred to a Foundation appointed third party mediator for arbitration.
+
+All decisions regarding removal of TSC members is subject to review by the body
+of Node.js collaborators. Should no fewer than one-quarter of the current
+Node.js project Collaborators (as defined by the TSC's governance and
+contribution policies) disagree with the TSC vote, the matter shall be referred
```
Trott: Is it your expectation that that means "collaborators" will be the 100+ members of the collaborators team? Or any Node.js GitHub repo, which could amount to more like 500 people?

jasnell: I've got no expectation currently. What do you think would be best?

gibfahn: As written I think `Node.js project Collaborators` ⇒ `the 500+ Org Collaborators` (rather than just Core Collaborators). +1 to it meaning Org Collaborators.

jasnell: If we say all org collaborators we would need to have a mechanism that allows 1/4 of those to register objections which can be difficult. That's not too say we shouldn't do this, we just need a workable process

stephenburgess8:
I'd like to propose a way to hold members of TSC accountable for following policy that would be a step before talking about removing them. There are many scenarios where a situation could merit some response, but maybe removal would be too extreme. In those cases, currently there are no consequences, which feels inadequate when dealing with points of contention. A third-party mediator wouldn't necessarily know how to best reprimand a member of TSC or a Working Group for failing to follow policy. I think if you offered a documented method to reprimand a member without removing them, it could address conflicts in an early stage rather than by the time it comes to asking someone to resign.

Here is a very rough example of how you could write a policy like that:

>A member of the steering committee (TSC) may bring a resolution for debate which proposes to censure another member of TSC or of a Working Group. This response can be taken when the person bringing the complaint, or the person representing someone with a complaint, can show that the member in question failed to follow Node policies.

>When a resolution is voted on, limitations on that member’s responsibilities may be included as part of the resolution. These limitations should be in response and in proportion to the issue at hand. Limitations of responsibility could be open for amendment in debate.

> An example of putting limitations on a member's responsibilities would be asking a member not to participate in moderation activities in Node projects for a period of six months. When a resolution expires, the limitations on their responsibilities would also expire.

>If a member receives multiple (3+) resolutions censuring them in a short period, their involvement in the project could be brought up for discussion.

I’d like to propose this because it seems like there should be a way to hold people accountable for following policy in a progressive way. Failing to follow the policies in a minor way could result in a minor censure without any limitations. Then, if something more serious happens there would be a way to address it through limiting someone's responsibilities for a while, rather than saying they can no longer participate at all. Limiting responsibilities could also be a method used by the third-party mediator. Another way to limit responsibilities would be requesting that a member not participate in discussion for the next two meetings. Whatever is most appropriate for whatever happened.

Thanks for reading my proposal. I hope it is helpful and not distracting. Also, thank you for working hard and building an awesome community.

TSC-Charter.md
jasnell: Actually, I've removed this entire bit. The Working Group based review process defined in https://github.com/nodejs/TSC/pull/319 would provide adequate coverage of this requirement.

jasnell:
@stephenburgess8 ... the type of thing you are describing is already possible under out current governance and would not be necessary to capture within the charter.

mhdawson:
@jasnell @stephenburgess8 while it may not need to be captured within the charter, I do think capturing the different types of complaints/actions. Not %100 sure where that documentation should end up but https://github.com/nodejs/TSC/issues/311 may be a good place to discuss it.

TSC-Charter.md
mhdawson: I think that makes it harder on those that want to abstain.  If you vote then you are less identifiable.  I see that in a vote you can vote yes, no or abstain and we don't identify who voted yes or no but we are going to identify to abstained.

jasnell:
Ping @nodejs/tsc members... just a reminder to please review these proposed charter changes.
@nodejs/ctc members... given the possibility of merging the CTC and TSC, I invite all of you to review these changes also and weigh in. Ideally we would have a simple majority of members signing off but as is our documented policy, if there are no objections to the proposed changes within a reasonable time frame, I will assume there is consent and will bring these proposed changes to the Board for their review and approval.

TSC-Charter.md
```diff
@@ -176,10 +187,42 @@ discussion will continue.
...
 Foundation board.
 
+### Votes to Remove Members
+
+Should a vote to remove a member result in a tie (50%/50% split of non
+abstaining participants) the matter shall be referred to a Foundation
+appointed independent third party mediator for arbitration.
+
+At any point during this process, the individual member in question, the
+individuals opening the request, or the TSC Chair may request that the issue
```
evanlucas: Does this mean that even after a vote, the TSC Chair or the individual member can still request this?

TSC-Charter.md
```diff
@@ -216,6 +259,26 @@ TSC.
...
 represent the Core Project on the TSC.
 
+## Section 11. Escalation of Disputes and Code of Conduct Violations
+
+Participation in the Node.js project is governed by a Code of Conduct policy
+that is established by the TSC. The TSC is required to establish a policy
+for the enforcement of Code of Conduct issues. Any reports of Code of Conduct
+or policy violations on the part of TSC members that are not resolvable through
+that established policy will be referred to binding, independent third party
+arbitration with a Foundation appointed mediator All TSC members, upon
```
evanlucas: missing period before "All"?

TSC-Charter.md
jasnell: As currently written, yes. But that's not likely what we want so I'm open to suggestions on how to rework this.

evanlucas: Yea, I don't agree with this part. With this left in, the actual TSC vote can be undermined. I almost think this entire paragraph should be removed.

jasnell: Well, I'd like to have a process by which the matter can at least be referred to mediation before the vote.

TSC-Charter.md
mhdawson: I'd be ok if it was Working Group or Team.

TSC-Charter.md
jasnell: I'll work on reworking this

jasnell: @evanlucas ... ok, please take a look now.

TSC-Charter.md
```diff
@@ -176,10 +187,42 @@ discussion will continue.
...
+### Votes to Remove Members
+
+Should a vote to remove a member result in a tie (50%/50% split of non
+abstaining participants) the matter shall be referred to a Foundation
+appointed independent third party mediator for arbitration.
+
+At any point prior to calling for a vote, the individual member in question, the
+individuals opening the request, or the TSC Chair may request that the issue
+be referred to a Foundation appointed third party mediator for
+arbitration.
```
isaacs: (comment copied so as to not get lost in the outdated fold)

It says that anyone can request that the issue be handled by a third party mediator instead of by a vote of the TSC, but it's not clear how or if that request is handled.

For example, if I'm on the TSC, but acting like a jackass, I might know that the members of the TSC are going to vote me out. In order to undermine this process, I request that it be referred to a third-party mediator, specifically because that gives me more leverage.

Does the foundation board get to say "No, that request is in bad faith, the TSC is going to vote on this"? And by what criteria would the foundation board make that decision?

I'd suggest removing this loophole. The conditions that lead to mediation should be objective and clear, and not dependent on anyone's request. Either it should always go to mediation, or only in the case of a tie, or never.

jasnell: Yeah, I've been stewing on that. The ideal is that it wouldn't have to go to mediation at all. It should also be a goal to avoid having to vote. So perhaps it should be simply if consensus cannot be reached, the matter goes to mediation. The challenge with that is that it encourages members to not take a position, e.g. I don't want to deal with this so abstaining from a clear position forces mediation. I don't know how to solve this yet.

TSC-Charter.md
```diff
@@ -36,16 +36,27 @@ approved by the Board.
...
-TSC memberships are not time-limited. There is no maximum size of the TSC.
-The size is expected to vary in order to ensure adequate coverage of important
-areas of expertise, balanced with the ability to make decisions efficiently.
-The TSC must have at least three members.
+TSC memberships are for one year terms that must be recertified. Motions for
+recertification are automatic and require a vote. Members are recertified if a
+simple majority of TSC members vote in favor of recertifying the individual as
+a TSC member. Such members will continue through the next year's term.
+Individuals failing recertification may return to the TSC after no less than
+six months following normal TSC motion.
```
jasnell: A tie would be less than simple majority so the individual would not be recertified. 

No, their vote for themselves wouldn't count. I can clarify that.

jasnell:
All.. I have asked the Foundation Board, and in particular Foundation Legal, to help clarify the language around mediation/arbitration. The goal is to get someone far more knowledgeable than I to help craft some language there that makes sense.

TSC-Charter.md
```diff
@@ -176,10 +187,42 @@ discussion will continue.
...
+Should a vote to remove a member result in a tie (50%/50% split of non
+abstaining participants) the matter shall be referred to a Foundation
+appointed independent third party mediator for arbitration.
+
+At any point prior to calling for a vote, the individual member in question, the
+individuals opening the request, or the TSC Chair may request that the issue
+be referred to a Foundation appointed third party mediator for
+arbitration.
+
+All decisions regarding removal of TSC members is subject to review by the
```
dankohn: s/`is`/`are`/

jasnell:
@mhdawson, can I please ask you to take this pr over?

TSC-Charter.md
hackygolucky: I will try to drudge up some language/process for this and coordinate with legal on how this needs to work.

mhdawson: @hackygolucky thanks that will be very helpful.

mhdawson:
@hackygolucky if you can share the language here, that might give us a better idea if it belongs in this PR or not.

jasnell:
Ping @nodejs/tsc ... still looking for more input on this. I have updated some of the proposed language.

mcollina:
This is a huge LGTM.

TSC-Charter.md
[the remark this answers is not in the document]
Fishrock123: Please do.

TSC-Charter.md
```diff
@@ -36,16 +36,27 @@ approved by the Board.
 
 ## Section 4. Establishment of the TSC.
 
-TSC memberships are not time-limited. There is no maximum size of the TSC.
-The size is expected to vary in order to ensure adequate coverage of important
-areas of expertise, balanced with the ability to make decisions efficiently.
-The TSC must have at least three members.
+TSC memberships are for one year terms that must be recertified. Motions for
```
Fishrock123: One year from... date of original addition / last recertification?

Kimberly050174:
Ok

Kimberly050174:
Ok

TSC-Charter.md
jasnell: The intention is all at once, once per year.

TSC-Charter.md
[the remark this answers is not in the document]
dshaw: Is there precedent for one's own vote for oneself not counting in other processes? With most board voting the vote for oneself, if you have the right to vote, counts.

TSC-Charter.md
```diff
@@ -176,10 +187,43 @@ discussion will continue.
...
 Note that, in addition to requiring a simple majority vote of the TSC, all
 changes to this charter are also subject to approval from the Node.js
 Foundation board.
 
+### Votes to Remove Members
+
+Should a situation arise that considers the removal of a TSC member, an
+attempt should first be made to resolve the issue without requiring a vote.
+The TSC Chair may, at their discretion, request that the issue be referred
+to a Foundation appointed third-party mediator for arbitration.
```
dshaw: I believe this should be "for mediation and, potentially, arbitration."

TSC-Charter.md
```diff
@@ -176,10 +187,43 @@ discussion will continue.
...
+
+Failing all other possible resolutions, a simple majority vote of TSC members
+is required to remove a member. Should such a vote to result in a tie (50%/50%
```
dshaw: s/vote to result/vote result

TSC-Charter.md
```diff
@@ -216,6 +260,39 @@ TSC.
...
+twelve months. The suspension period must be specified in advance of the
+decision. The members privileges as a TSC member will be restored automatically
+after the given period has passed. The TSC may decide, through simple majority
+vote, to lift the suspension early.
+
+## Section 13: Selection of Mediators
+
+When it becomes necessary within the context of this charter to select a
+third party mediator for arbitration, a qualified mediator shall be selected by
+the Node.js Foundation Board of Directors (or an individual designated by the
```
dshaw: We refer to the Node.js Foundation as providing the mediator, but here we state that the board is responsible for selection. Introducing the board into this statement may be unnecessary. I recommend reviewing this bit with Foundation Legal.

TSC-Charter.md
Trott: > The intention is all at once, once per year.

I'm not going to expend a lot of effort trying to change this, but I'll say that I'd prefer it be rolling so that there is a rhythm. "What two or three  people need to be recertified this month? And who is up for recertification next month?" instead of "Oops, did we go two-and-a-half years without recertifying anyone again?" (Sort of what happened with the TSC Director/Chair position at last once, no?)

mhdawson:
@jasnell I think I've seen you jump back into a few of these issues, do you still want me to take it over or can you update based on comments ?

jasnell:
I'd prefer you to take this over. Feel free to push any further edits to my branch.

MylesBorins:
Hey All, is this ready to bring to the board or does it require more work before we have consensus?

jasnell:
Definitely requires more work.

TSC-Charter.md
fhinkel: I would prefer 2 year terms.

TSC-Charter.md
```diff
@@ -176,10 +187,43 @@ discussion will continue.
 For all votes, a simple majority of all TSC members for, or against, the issue
 wins. A TSC member may choose to participate in any vote through abstention.
 
+While the results of all votes must be made public, the actual individual
+ballots cast for most votes may be made public or confidential at the discretion
+of the TSC Chair. However, individual ballots for recertification or removal of
```
fhinkel: I feel like one member wanting it to be confidential should be enough, not depending on the chair.

TSC-Charter.md
```diff
@@ -176,10 +187,43 @@ discussion will continue.
...
+TSC members must remain confidential. The names of members voting or abstaining
+in all votes must be made public with a clear indication given of whether the
+individual voted or abstained.
```
fhinkel: Why is abstaining public? Not sure I like that.

TSC-Charter.md
fhinkel: Should this be 66%/33% or do we like the potential almost 50% non-support?

TSC-Charter.md
```diff
@@ -216,6 +260,39 @@ TSC.
 * **Maintainer**: a Collaborator within a Core Project elected to
 represent the Core Project on the TSC.
 
+## Section 11. Escalation of Disputes and Code of Conduct Violations
+
+Participation in the Node.js project is governed by a Code of Conduct policy
+that is established by the TSC. The TSC is required to establish a policy
```
fhinkel: Why is this not established by CommComm?

mhdawson:
With other proposals like https://github.com/nodejs/TSC/issues/382 I'm not sure if this PR is still applicable.

MylesBorins:
+1 to closing this if @jasnell agrees

jasnell:
I'm not working on this further. Close if you'd like

Trott:
Closing. Re-open if that's a mistake.

PR #319, opened 2017-08-22
meta: add ability for working groups to override the TSC

jasnell:
this is a proposal to add a formal process allowing the chartered working groups to vacate TSC decisions.

@nodejs/tsc @nodejs/ctc

### DIFF
--- a/WORKING_GROUPS.md
+++ b/WORKING_GROUPS.md
@@ -25,12 +25,12 @@ Chartered Working Group vote to challenge a TSC decision, the issue will be
 raised automatically for consideration by all chartered Working Groups. Should
 a simple majority of all chartered Working Groups vote with a simple majority
 of their documented members to overturn the TSC decision, the decision will be
-vacated and returned the TSC for further consideration.
+vacated and returned to the TSC for further consideration.
 
 Individual Working Groups may choose to abstain from participation in any
 such action.
 
-Working Groups will decide for themselves what process to be followed for
+Working Groups will decide for themselves what process will be followed for
 organizing and conducting such votes.
 
 ## Process:
--- a/WORKING_GROUPS.md
+++ b/WORKING_GROUPS.md
@@ -33,6 +33,15 @@ such action.
 Working Groups will decide for themselves what process will be followed for
 organizing and conducting such votes.
 
+Note that the simple majority vote excludes those who explicitly abstain. For
+instance, if a Working Group has ten members, and two of those members
+abstain, then five of those Working Group members constitute a simple majority
+of members.
+
+Likewise, if there are ten chartered Working Groups, and two of those choose to
+abstain, then five of those Working Groups constitute a simple majority of
+Working Groups.
+
 ## Process:
 
 * [Starting a Working Group](#starting-a-wg)
@@ -104,7 +113,7 @@ ratified by that WG.
 You can submit the WG charter for ratification by sending
 a Pull Request to this document, which adds it to the
 list of current Working Groups. Once ratified the list of
-members should be maintained in the Working Group's
+members must be maintained in the Working Group's
 README.
 
 ## Bootstrap Governance
--- a/WORKING_GROUPS.md
+++ b/WORKING_GROUPS.md
@@ -17,6 +17,22 @@ should be dissolved and the responsibility for governance absorbed back into
 the TSC. A Working Group can be dissolved either through consensus of the
 Working Group membership or normal TSC motion and vote.
 
+## Working Group Oversight of TSC Actions
+
+Chartered Working Groups are granted an oversight capability over all TSC
+decisions. Should a simple majority of the documented members of any single
+Chartered Working Group vote to challenge a TSC decision, the issue will be
+raised automatically for consideration by all chartered Working Groups. Should
+a simple majority of all chartered Working Groups vote with a simple majority
+of their documented members to overturn the TSC decision, the decision will be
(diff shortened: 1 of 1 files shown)


MylesBorins:
I am very much in favor of this.

Fishrock123:
Contacting everyone would be difficult but hopefully rare.

I'm uncertain if it should or shouldn't return to the TSC though

jasnell:
It should. Essentially this is the WGs telling the TSC strongly that it made a wrong decision and needs to reconsider. Once the TSC and CTC merge back into a single body, this should be fine.

gr2m:
This would be very helpful as an official way to hold members of the TSC accountable for their decisions. I agree that the bar is set high but it should be the exception. It is an improvement of the current situation and I’d ask we merge it in and continue to iterate

jasnell:
Note that this cannot be merged without TSC/CTC sign off.

WORKING_GROUPS.md
```diff
@@ -17,6 +17,22 @@ should be dissolved and the responsibility for governance absorbed back into
...
 
+## Working Group Oversight of TSC Actions
+
+Chartered Working Groups are granted an oversight capability over all TSC
+decisions. Should a simple majority of the documented members of any single
+Chartered Working Group vote to challenge a TSC decision, the issue will be
+raised automatically for consideration by all chartered Working Groups. Should
+a simple majority of all chartered Working Groups vote with a simple majority
+of their documented members to overturn the TSC decision, the decision will be
+vacated and returned the TSC for further consideration.
```
Trott: Micro-nit: `returned the TSC` -> `returned to the TSC`

WORKING_GROUPS.md
```diff
@@ -17,6 +17,22 @@ should be dissolved and the responsibility for governance absorbed back into
...
+
+Individual Working Groups may choose to abstain from participation in any
+such action.
+
+Working Groups will decide for themselves what process to be followed for
```
Trott: Micro-nit: `process to be followed` -> `process will be followed`

Trott:
LGTM (but I'm avoiding the official green check because this is the TSC repo and I"m not TSC).

Trott:
There may be an issue where we need a vote from a few more WGs to come to a conclusion and they take a looooong time to get it together to have the vote.

I'm OK with that. To me, that basically means _status quo_ (the TSC decision stands) until enough WGs vote to vacate it or support it.

However, if anyone else reads this differently or if that is not the intended outcome of such an event, then it should be clarified in the text.

jasnell:
@nodejs/tsc and @nodejs/ctc ... please review. This is a fairly major policy change. Ideally we will have a simple majority of members signing off but, as is the documented policy, if there are no objections after a reasonable period of time I will assume consent and will get this landed.

Trott:
Idle though, might not need addressing, but might be worth considering. Not sure. Here we go:

What happens if there's an infinite loop?

* TSC passes Proposal X
* WGs reject it, so Proposal X is vacated and returned to the TSC
* TSC passes Proposal X again...
* WGs reject it again...

Is that OK? Or does there need to be a "nope, sorry, this decision is final"? Like, in the US, if the federal legislature passes a law, and the executive branch vetoes (rejects) the law, the legislature can override the veto by holding a vote and passing the legislation with a 2/3 majority.

Trott:
Totally pedantic nit: Since presumably the only votes are "yes" or "no", it doesn't really make much sense to refer to a `simple majority`. It's just `a majority`. 

(A `simple majority` is in contrast to an `absolute majority` but that's only meaningful if there are three or more choices.)

Trott:
This seems OK to me, but I'm not sure if I'm in favor of it or not. No objection, but I'm going to refrain from giving it a green checkmark, at least for now. For me, a lot depends on how much we require WGs to truly be active going forward, as well as being rigorous about chartering informal WGs that are highly active.

Something else that might also affect my opinion: Are i18n teams chartered? Is there one big i18n group or is each team a separate WG?

jasnell:
If it goes into an infinite loop we have more systemic problems to deal with. I'd say lets not worry about that for now.

The i18n working groups are not independently chartered as far as I know.

bnoordhuis:
I don't know how I feel about this.  I'm not an authoritarian but I think it's right and proper that the TSC has the final responsibility and the final say on matters.

This change opens up the possibility of perpetual stalemate.  James seems to think it's academic.  It very well might be but I don't really want to go there and find out we were wrong.

Rich is right to point out that this gives Working Groups power that may be out of proportion to their activity and relevance.

There is potential for abuse.  It takes little effort for a WG to challenge a TSC decision but it's a large effort to collect and discuss the input from all the other WGs.

I'll err on the side of caution and -1 this for now.  Clear downsides, unclear upsides.

<sup>Caveat emptor: I'll be away for a few days and probably won't get around to replying until next week.</sup>

mcollina:
How about we raise the requirement to two working groups? I think that would remove completely the possibility for abuse.

jasnell:
I spoke to the board about this also and the feedback was that the "override" part was likely problematic. The suggestion was that an "appeal" process would likely be better. We should just take another pass at this

jasnell:
@mhdawson, can I please ask you to take this pr over?

jasnell:
I am withdrawing this PR. I do believe that some mechanism here is required, but the feedback from the board was that the approach suggested in this PR was not ideal.

PR #341, opened 2017-09-09
Moderation Policy: Use consistent terminology

Trott:
Bans are definted as _temporary_ or _indefinite_, but then usage later
on is all about _temporary_ vs. _permanent_. This change makes the
terminology consistent throughout the policy.

Additionally, it re-orders some text so that _temporary_ always appears
before _indefinite_.

### DIFF
--- a/Moderation-Policy.md
+++ b/Moderation-Policy.md
@@ -164,14 +164,14 @@ a Post from Moderation.
   * Identify all changes made to the Post, and
   * Identify the steps previously taken to resolve the issue.
   * If the Moderation action included Banning, an indication of whether the Ban
-    is permanent or temporary is required, along with an issue posted to the
+    is temporary or indefinite is required, along with an issue posted to the
     moderation repository justifying the action.
 * Explanations of Moderation actions on Collaborator Posts must be provided in:
   * A new post within the original thread, or
   * A new issue within the private nodejs/moderation repository.
 * Any Collaborator habitually violating the Code of Conduct or this Moderation
   policy may be Banned temporarily or, in extreme cases, Removed and Banned
-  permanently.
+  indefinitely.
 
 ### Non-Collaborator Posts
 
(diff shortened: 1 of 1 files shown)


Trott:
@nodejs/tsc  @hackygolucky, @bnb, @gibfahn, @benjamingr @refack @ryanmurakami @othiym23

mscdex:
I wonder if we should be careful about word choice here. To me, 'permanent' implies that there were *never* be any chance of the ban being lifted, whereas 'indefinite' merely implies there is no defined duration of the ban, so it *could* be lifted at any point in the future.

Trott:
> I wonder if we should be careful about word choice here. 

I suspect _indefinite_ was initially used to signal that something isn't completely irreversible. One upside is that it may mean we're more likely to actually use an indefinite/permanent ban if it's called _indefinite_. There may be a psychological resistance against doing something "permanently" that isn't there against doing something "indefinitely" even though they amount to the same thing in some contexts.

mscdex:
Right, but I'm just saying it's important that we all agree on what meaning/wording we really want and not just pass this off as a simple consistency-seeking PR.

benjamingr:
LGTM

ofrobots:
Indefinite seems like a better word choice. LGTM.

gibfahn:
Agree with @mscdex that this is a decision, and +1 for indefinite, I think that more accurately reflects most bans.

MylesBorins:
+1 for indefinite

Moderation-Policy.md
```diff
@@ -217,14 +217,14 @@ decision was made should be added to the moderation repository thread.
...
-An Indefinite Ban is set for an unspecified period of time and may only be
-lifted for an individual through a simple majority vote of the the Moderation
-Team.
-
 A Temporary Ban is time limited, with the timeframe decided on by the Moderation
 Team at the time of issuing, depending on the severity of the issue. Recommended
 default options are 24-hour, 48-hour, and 7-day periods.
 
+An Indefinite Ban is set for an unspecified period of time and may only be
+lifted for an individual through a simple majority vote of the the Moderation
```
thefourtheye: Nit: There are two "the"s here. I see that it was already there, but now that it is moved around, I think we can do this small change.

jasnell:
Landing given the lack of objections

PR #391, opened 2017-10-21
Open communications should be a priority

koresar:
As per discussion with some members of the board here https://github.com/nodejs/board/issues/70#issuecomment-328195328 I would like to suggest the following change to the **Guiding Principle** as proposed by @williamkapke.

Let's copy the best.

### DIFF
--- a/TSC-Charter.md
+++ b/TSC-Charter.md
@@ -6,6 +6,23 @@ The Node.js Foundation will operate transparently, openly,
 collaboratively, and ethically. Project proposals, timelines, and status
 must not merely be open, but also easily visible to outsiders.
 
+We endeavour to conduct as much discussion in public as possible. This 
+encourages openness, provides a public record, and stimulates the 
+broader community.
+
+However sometimes internal private communication is necessary. You must 
+never divulge such information in public without the express permission 
+of the discussion participants. Also never copy a message between 
+private and public comms. Such an event would go beyond the normal need
+for communication ettiquette and be a serious breach of confidence. It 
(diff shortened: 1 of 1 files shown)


koresar:
Could you please propose the place where this would be more suitable?

jasnell:
The expectations doc being worked on in the admin repo would likely be more appropriate

koresar:
Huh. Funny. That's what my change is about. Open (not for writing maybe)
any communication. Including the "admin" and all the other closed repos.

MylesBorins:
The admin repo is an open repo. The only closed repos in the org is
moderation and security

koresar:
I see. Thank you for the info. I've just found the repo.
Do you think it's reasonable to move this PR change to this file?
https://github.com/nodejs/TSC/blob/master/CODE_OF_CONDUCT.md

bnb:
As @jasnell said, I think the https://github.com/nodejs/admin repo is an entirely appropriate place. I don't think the CoC is an appropriate place for discussion which medium of communications we use, as it will detract from the other parts of the message in that doc.

mhdawson:
I suggest you take the content of this PR, and submit it as a document titled "Communications-policy.md" in the admin repo (https://github.com/nodejs/admin/pull/12).  I'm also happy to do that if you'd prefer.

koresar:
That's neat suggestion. Please don't waste your time please. Let me do it.

Thank you for the help.

MylesBorins:
@koresar I'm going to go ahead and close this as it does not appear to be actionable in the current state. Please lmk if you want support with anything in the future

PR #394, opened 2017-10-23
Moderation Policy: no bots but our bots

Trott:
ping @nodejs/tsc @nodejs/moderation

### DIFF
--- a/Moderation-Policy.md
+++ b/Moderation-Policy.md
@@ -196,6 +196,8 @@ of the [Code of Conduct][].
   intention to collaborate in good faith, swift actions may be taken without
   following the above procedures including: removing Posts, Banning
   indefinitely, and reporting accounts to GitHub.
+* Accounts that are reasonably believed to be bots (other than bots authorized
+  by the TSC) are subject to immediate Banning.
 
 Note that Moderating non-Collaborator posts can often lead to retaliation or
 escalation of inappropriate behavior by the individual whose post is being


MylesBorins:
@Trott we may want to do this in the admin repo

The Moderation doc is living in two places. We likely should move this doc over there first

vsemozhetbyt:
Refs: https://github.com/nodejs/moderation/issues/155

othiym23:
@MylesBorins It's [already there](https://github.com/nodejs/admin/commit/527334725aa66e9c6b218239047a8152c199afeb) (I only figured that out on Friday myself, don't feel bad). @Trott -- you want to close this and move the proposal there? And should we start the process of removing the moderation policy here? It's at best confusing having it in two places, even if one of them is a deprecated repository.

othiym23:
Also, on the proposal's merits: I think this is a sensible proposal, and also would be interested in basically codifying a "quick to ban, quick to forgive" policy around bots or any other behavior that can rapidly trigger events in repos, because it's so easy for bot-like activity to create a lot of work for a lot of people.

MylesBorins:
I've opened https://github.com/nodejs/TSC/pull/395

MylesBorins:
This is going to have to be moved to the admin repo

Trott:
> This is going to have to be moved to the admin repo.

https://github.com/nodejs/admin/pull/31

PR #395, opened 2017-10-23
doc: move Code of Conduct and Moderation Policy to admin repo

MylesBorins:
This will require https://github.com/nodejs/admin/pull/25 to land
first


MylesBorins:
I've updated this to move both the CoC and Moderation Policy

MylesBorins:
I've also updated the base policies that are copied when creating new working groups

jasnell:
I think I'm generally -1 on moving the CoC. I think there is value in having the CoC here and there's no harm in the CommComm having it's own copy or even it's own text.

I'm good with moving the Moderation policy.

MylesBorins:
@jasnell can you expand on the specific value of the TSC having it's own CoC? It seems odd that the CoC would live in this repo and the guidelines in another, pretty counter intuitive.

The CommComm will need a CoC as well, and I think it is a good thing, at least for the beginning, that we coalesce.

jasnell:
> can you expand on the specific value ...

Every time we move this document back a layer, we essentially add a new layer of abstraction. Having the TSC own the CoC for the technical project makes the TSC more directly responsible for ensuring it's enforcement within the project. Yes, there are some logistical processes and activities that we can share effectively with CommComm (like the day to day management of the github org, moderation activities, etc), but the idea of the TSC actively *managing* the technical project and actively enforcing it's own policies is something that cannot be outsourced or shared.

Yes, it would be a good thing for the CommComm to have a CoC, and ideally that would use the same text as the TSC CoC, but they don't have to live in the same spot.

mhdawson:
I don't think we need separate copies unless they are different.  

Could just just have a line in the README.md for the TSC that states the CoC is the version in the admin repo.  If at some point the TSC one needs to diverge the different copy could be in the TSC repo and the reference could be changed.  I don't think having a duplicate copy is going to change the level of ownership.

MylesBorins:
@mhdawson that was my thought as well. Unless we explicitly need to diverge I think having separate copies only increases the odds of divergence. Further, if both TSC + commcomm are using the the same CoC, it doesn't really make sense to have it anywhere but admin

MylesBorins:
I've updated the copy on the code of conduct
@jasnell can you ptal?

PR #408, opened 2017-11-01
remove mentor program: obsolete

Trott:
AFAIK, we don't do the mentor thing for projects. We can always add this back if it's something we start putting some structure and energy into, but for now, remove it.


jasnell:
We haven't actively done it in a while because there are no projects that need it. I'm neutral on this change.

PR #421, opened 2017-11-17
remove "Core Projects" and "Maintainer" from charter

Trott:
Core Projects are not defined anywhere in the Charter. As far as I know,
we do not have Core Project representation as defined in the Charter. We
do not intend to expand the number of Core Projects, and I'm not even
sure we have any at the current time. Remove this material from the
Charter for clarity and concision. (This will require board approval.
Technically doesn't require TSC approval, but that would be nice.)

Removing *Core Projects* from the Charter makes the *Maintainer* terminology superfluous, so that is being removed too. The *Maintainer* terminology is not often used in reference to Node.js in my experience, and when it *is* used. it is synonymous with *Collaborator* and in contradiction with the definition in the Charter. So, again, this removes the *Maintainer* terminology too.

### DIFF
--- a/TSC-Charter.md
+++ b/TSC-Charter.md
@@ -10,9 +10,8 @@ must not merely be open, but also easily visible to outsiders.
 
 Most large, complex open source communities have both a business and a
 technical governance model. Node.js Foundation’s technical leadership
-contains both a Technical Steering Committee (“TSC”) and Maintainers for
-major components or subsystems. Node.js Foundation’s business leadership
-is instantiated in a Board of Directors (the “Board”).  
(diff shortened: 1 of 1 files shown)


MylesBorins:
This will need to be approved by the board.

Next meeting is Nov 27

Trott:
@MylesBorins Did this get board approval?

MylesBorins:
@Trott with the timing I missed getting this on the agenda, I will get this sorted in the December meeting.

Trott:
Approved unanimously at the Board meeting December 18 2017. Landing! 🎉

Issue #353, opened 2017-09-15
Decharter Intl WG


jasnell:
I would like to propose dechartering the Intl WG in favor of it being a more informal Team. I'm not convinced that we ever really had much benefit of it being a formally chartered working group. A team would still be just as effective without having any of the additional overhead of a chartered working group.

ping @srl295

fhinkel:
ping @srl295

mhdawson:
@srl295 unless you object I'd agree that Intl could be a team.

srl295:
sure— sounds OK off the top of my head (and I'll go change my slides…)

srl295:
^ I could not find a lot of details on team vs. WG. I filed the above issue nodejs/Intl#45 over any changes that should happen to the Intl repo. In particular, does the Intl _repo_ need to close, because there are some valuable issues there— but, they could move to issues in other repos but with appropriate tags.

Trott:
@srl295 Team = GitHub team, so @nodejs/intl would be the team. Yeah, not a lot of documentation on how that differs from a chartered WG. We could definitely improve there. The repo does not need to close. (Testing WG was de-chartered but the repo remains, for example.)

srl295:
What's the next step here?

MylesBorins:
Ping @nodejs/tsc

Trott:
> Ping @nodejs/tsc

https://github.com/nodejs/nodejs.org/pull/1513
https://github.com/nodejs/TSC/pull/437

Trott:
Going through remaining open issues in https://github.com/nodejs/Intl/issues right now. Planning on archiving the repo (basically making it read-only) sooner or later.

mhdawson:
@srl295 earlier discussion was that the repo did not necessarily need to close if the WG became a team.  Is the repo still useful for the team ?

Trott:
@mhdawson @srl295 OK, I won't archive the repo for now and I'll stop my march through all the issues to close them that I started a few minutes ago. It does seem that the Intl repo is a bit of a graveyard though. I think the issues in that repo would probably get more activity/visibility elsewhere.

jasnell:
The discussion appears to have stalled. What is the next step here?

srl295:
@jasnell did it? What's left to do? I have a PR https://github.com/nodejs/Intl/pull/47 to redirect to i18n, and I'm going to go over the rest of the Intl issues and refile…

jasnell:
[...]

PR #437, opened 2017-12-11
remove dechartered Intl working group

Trott:
Ref: https://github.com/nodejs/TSC/issues/353

### DIFF
--- a/README.md
+++ b/README.md
@@ -233,7 +233,7 @@ For more information look here:
 * [Working Groups](WORKING_GROUPS.md)
 * Top-Level Projects
  * Core TLP
-  * Core WGs (streams, http, Intl)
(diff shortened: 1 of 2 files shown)


maclover7:
The readme has a reference to the nodejs/intl repo that should also probably be changed: https://github.com/nodejs/TSC#adjacent-projects

Trott:
> The readme has a reference to the nodejs/intl repo that should also probably be changed: 

I left the reference to the repo (since it's not going anywhere for now) but removed the reference to the Intl WG. There was also a stray reference to the http WG which is also defunct.  Oof, and CTC and Testing WG.... I'm not going to remove all those, but instead propose removing the list from the README. I don't think it gets used and it clearly doesn't get maintained....

PR #456, opened 2017-12-27
doc: update GitHub Org Management Policy to reflect actual practice

Trott:
There are several aspects of the GitHub org management policy that are
not followed in practice and may be unworkable. This change tries to
move the document closer to actual practice.

@nodejs/moderation @nodejs/community-committee @nodejs/tsc 

Note that the intention here isn't to say that we *can't* or *shouldn't* do the things that are in the policy currently, but since we don't and since it's clear from https://github.com/nodejs/admin/issues/33 that we aren't going to do so imminently, let's update the policy to be more accurate, and then discuss any changes that people want.

I know this will be disappointing and frustrating to many people, but I also think it's better to have an accurate document for us to discuss than to have it be _de facto_ acceptable that we simply ignore this document.


GitHub-Org-Management-Policy.md
```diff
@@ -1,31 +1,23 @@
...
-Community Committee Chair, and Node.js Foundation Education Manager shall be
-the only individuals granted Owner permissions within the Node.js GitHub
-Organization. Should, at any point in the future, the Node.js Foundation
-Board establish a Community Committee Director position equivalent to the
-TSC Director, that individual would also be automatically granted Owner
-permissions within the organization.
-
-Upon the completion of the terms of service for each of these individuals,
-their Owner permissions within the GitHub organization shall be removed.
+The TSC Director shall be the only individuals granted Owner permissions within
```
addaleax: s/TSC Director/TSC members/ (for current practice) or s/individuals/individual/?

pup: `individuals` should be singular

GitHub-Org-Management-Policy.md
```diff
@@ -48,22 +40,6 @@ Any repository created under the Node.js GitHub Organization is considered to be
...
-All repositories must have a README that clearly identifies the purpose of the
-repository, governance documentation that details how the repository is managed,
-and an indication of whether the repository falls under the operational
-direction of either the TSC or CommComm. 
-
-Repositories that are under the operational direction of the TSC are subject to
-TSC oversight.
-
-Repositories that are under the operational direction of the Community Committee
-are subject to CommComm oversight.
```
addaleax: It’s not quite obvious to me why this is being removed … because we don’t currently follow it?

GitHub-Org-Management-Policy.md
pup: oh, i wasn't sure whether this _actually_ only meant TSC director, if it means the TSC members then my review is redundant

GitHub-Org-Management-Policy.md
Trott: Sort of. Also, it's redundant (or maybe just has some nuance I'm missing?). It basically can be paraphrased as: "Repositories that are under CommComm oversight are subject to CommComm oversight." Like, I read that, and other than including CommComm, I'm not sure why it's even there.

GitHub-Org-Management-Policy.md
Trott: Fixed, thanks.

Trott: Thanks. I removed `Director`. (Missed it the first time.) Should say: The TSC shall be the only individuals granted Owner permissions....

(Note that I'm not saying this is the way it *should* be but it is the way it *is* and there's at least some resistance to changing it. Which reminds me, that I should add a Ref in one of these comments to the issue where this is being discussed.)

rvagg:
I'm not a fan of this doc in general but this is an improvement so +1 from me

GitHub-Org-Management-Policy.md
```diff
@@ -1,31 +1,23 @@
...
-Community Committee Chair, and Node.js Foundation Education Manager shall be
-the only individuals granted Owner permissions within the Node.js GitHub
-Organization. Should, at any point in the future, the Node.js Foundation
-Board establish a Community Committee Director position equivalent to the
-TSC Director, that individual would also be automatically granted Owner
-permissions within the organization.
-
-Upon the completion of the terms of service for each of these individuals,
-their Owner permissions within the GitHub organization shall be removed.
+The TSC shall be the only individuals granted Owner permissions within the
```
mcollina: I think it should be “The TSC members”.

GitHub-Org-Management-Policy.md
```diff
@@ -48,22 +40,6 @@ Any repository created under the Node.js GitHub Organization is considered to be
 a project under the ownership of the Node.js Foundation, and thereby subject
 to the Intellectual Property and Governance policies of the Foundation.
 
-No repository may be deleted, transferred in to, or transferred out of, the
-Node.js Foundation GitHub Organization without a simple majority of both
-the TSC and CommComm in *favor* of the action. In certain cases, Node.js
-Foundation Board of Directors approval may also be required.
```
mcollina: I think this paragraph is still needed, without the CommComm reference.

GitHub-Org-Management-Policy.md
Trott: Thanks, fixed.

GitHub-Org-Management-Policy.md
Trott: Reinstated. Thanks.

GitHub-Org-Management-Policy.md
mhdawson: I agree its bad to have a doc that does not reflect reality but I think instead of removing references to the Community Committee having oversight as the interim solution, we should be making the CommComm chairs owners in addition to all TSE members.  I don't think its our final state but I think its a more acceptable interim state than simply removing any CommComm onership.

Trott: @mhdawson If you would like to add a commit to make that change, feel free. We will need everyone who has approved to re-approve, I think (or remove their approval if they don't approve). My intention had been to open a PR to that effect immediately after this lands, but if you think it's better to do that now, I'm certainly OK with seeing what everyone thinks about that.

mhdawson:
Pushed commit with my suggested changes after discussing with @Trott .

@mcollina @fhinkel @cjihrig can you re-review as my commit changes enough to warrant you reviewing again and either confirming you are ok with it or revoking your ok.

mhdawson:
@rvagg see you had +1 instead of approval with button so FYI to you to re-review as well.

GitHub-Org-Management-Policy.md
```diff
@@ -1,31 +1,23 @@
...
-Community Committee Chair, and Node.js Foundation Education Manager shall be
-the only individuals granted Owner permissions within the Node.js GitHub
-Organization. Should, at any point in the future, the Node.js Foundation
-Board establish a Community Committee Director position equivalent to the
-TSC Director, that individual would also be automatically granted Owner
-permissions within the organization.
-
-Upon the completion of the terms of service for each of these individuals,
-their Owner permissions within the GitHub organization shall be removed.
+TSC members and the Chair of the Community Comittee are the only individuals granted
```
Trott: _Committee_ is misspelled here.

Trott: (Actually, never mind, I'll correct the spelling. No need to make someone else do it when I can just use GitHub's edit feature and be done with it.)

cjihrig:
I'm still OK with this after @mhdawson's commit.

GitHub-Org-Management-Policy.md
```diff
@@ -0,0 +1,70 @@
...
+The [Node.js admin repository](https://github.com/nodejs/admin) will serve as
+the central location for managing Node.js GitHub Organization administrative
+activities. Only Node.js GitHub Organization owners, TSC members, and Community
+Committee members will have write permissions to the Node.js admin repository.
+
+## Organization Roles
+
+### Owners
+
+TSC members and the Chair of the Community Committee are the only individuals granted
```
pup: there can be multiple chairs of the commcomm

mhdawson: Agreed, I thought that was a detail we could leave out with the the assumption it would include chairs in the exception case (which of course is what we have now).

pup: that works!

GitHub-Org-Management-Policy.md
```diff
@@ -0,0 +1,70 @@
...
+provided as a development resource by the Node.js Foundation under the direction
+of the Node.js Technical Steering Committee (TSC) and Node.js Foundation
+Community Committee (CommComm).
+
+## Node.js Admin Repository
+
+The [Node.js admin repository](https://github.com/nodejs/admin) will serve as
+the central location for managing Node.js GitHub Organization administrative
+activities. Only Node.js GitHub Organization owners, TSC members, and Community
+Committee members will have write permissions to the Node.js admin repository.
```
joyeecheung: `s/will have/have/`?

GitHub-Org-Management-Policy.md
```diff
@@ -0,0 +1,70 @@
...
+
+### Owners
+
+TSC members and the Chair of the Community Committee are the only individuals granted
+Owner permissions within the Node.js GitHub Organization.
+
+### Members
+
+GitHub users are added as members to the Node.js GitHub Organization when they
+are added to any Working Group. Organization Owners should add new members
```
joyeecheung: At the moment we also add users when they are added to a team (not necessarily belong to any charted WG), so maybe `...to any Working Groups or teams.`?

GitHub-Org-Management-Policy.md
```diff
@@ -0,0 +1,70 @@
...
+to the organization when requested by a Working Group.
```
joyeecheung: `...by a Working Group or a team.`

Also it may be worth mentioning the owner should suggest new members to follow https://github.com/nodejs/admin/blob/master/CODE_OF_CONDUCT.md and enable 2FA prior to adding them to the organization.

GitHub-Org-Management-Policy.md
```diff
@@ -0,0 +1,70 @@
...
+creating the repository. Such requests must be posted as issues in the
+Node.js admin repository.
+
+Any repository created under the Node.js GitHub Organization is considered to be
+a project under the ownership of the Node.js Foundation, and thereby subject
+to the Intellectual Property and Governance policies of the Foundation.
+
+No repository may be deleted, transferred in to, or transferred out of the
+Node.js Foundation GitHub Organization without a simple majority of both the
+TSC and CommCom in favor of the action. In certain cases, Node.js Foundation
```
joyeecheung: `s/CommCom/CommComm`

GitHub-Org-Management-Policy.md
```diff
@@ -0,0 +1,70 @@
...
+
+### Members
+
+GitHub users are added as members to the Node.js GitHub Organization when they
+are added to any Working Group. Organization Owners should add new members
+to the organization when requested by a Working Group.
+
+## Repositories
+
+Any organization member may request the creation of a new repository within the
```
joyeecheung: Maybe moving how to do this to the first sentence to increase visibility, e.g. `...by opening an issue in the Node.js admin repository.`?

GitHub-Org-Management-Policy.md
```diff
@@ -0,0 +1,70 @@
...
+may occur at any time subject to the policies outlined in the Moderation
+Guidelines.
+
+## Use of Bots and Services
+
+Automation tools such as bots and third-party services on any repository must
+be approved by the TSC and CommComm and are subject to regular security audits.
+Bots that perform actions on behalf of the project (such as moderation or membership
+management actions) are required to maintain a log, accessible to all individuals
+granted Owner permissions, of all actions taken.
```
joyeecheung: In the current practice, only people who have added their public keys to the related directory there have access. I would consider documenting this as `.. required to maintain a log of all actions taken. The secrets and the logs of the bots and the services are accessible to individuals granted permissions in the nodejs/secrets repo`.

Also it might be worth mentioning https://github.com/nodejs/automation/blob/master/enable-travis-under-nodejs.md here, maybe something like `See [this document][] on how to migrate Travis integration of a repository transferred into the Node.js organization`.

GitHub-Org-Management-Policy.md
```diff
@@ -0,0 +1,70 @@
...
+each of the Technical Steering and Community Committees.  A simple majority of
+each group *rejecting* the creation of the repository is required to block
+creating the repository. Such requests must be posted as issues in the
+Node.js admin repository.
+
+Any repository created under the Node.js GitHub Organization is considered to be
+a project under the ownership of the Node.js Foundation, and thereby subject
+to the Intellectual Property and Governance policies of the Foundation.
+
+No repository may be deleted, transferred in to, or transferred out of the
```
joyeecheung: `s/in to/into/`

bnb:
@Trott Could you please add `cc-agenda` label to this issue? 🙏

GitHub-Org-Management-Policy.md
Trott: I'm not sure I'll get this right. Maybe you or someone else can propose this particular change subsequently? (Or add a squash commit to this PR if you prefer?)

Trott:
Per TSC meeting discussion, this is ready to land but we're going to leave it open for another 48 hours or so. If anyone has objections, now is the time to say so.

pup:
@Trott let's wait until after tomorrow's CC meeting

GitHub-Org-Management-Policy.md
joyeecheung: @Trott Sure, I can open a subsequent PR for this.

pup:
nobody from the commcomm objected to this, so i'd say it's ready to land @Trott

hackygolucky:
I'm +1 to shipping this from the CommComm and Moderation team side if we can immediately start working on a doc that enables both groups to do the work they need to do by getting appropriate permissions to the org. I understand the need for accurate reflection of the-world-as-it-exists.

Trott:
Now that this has landed, it is official policy that CommComm chairs have owner permissions, so I'm going to add them. (When is the next election for CommComm chair?)

pup:
@Trott september 2017

pup:
@hackygolucky do you want to create an issue that tries to pull all of the future and past discussion about this together?

PR #39, opened 2016-01-28
Add initial TLP application for express

jasnell:
The express project is applying to become an incubating top level project.
/cc @nodejs/tsc @Fishrock123 @nodejs/collaborators

PR #462, opened 2018-01-04
remove "top level projects"

Trott:
Top Level Projects feature prominently in the README but are not
actually a thing. There are currently zero Top Level Projects (TLPs)
excluding core. Seems like the usual path nowadays is for a project to
post in the admin repo to ask to be brought under the foundation.

So...simplify our README and remove gaps in information by removing
TLP text.


Trott:
> but I do believe Express is still under the foundation and this may be the only place where that is documented.

@mhdawson When you say "Express is...under the foundation", what exactly does that mean?

mhdawson:
Express was brought under the auspices of the Node.js foundation. Look at "Applications/Express.md".  A lot of that happened in terms of formation.  Look here https://expressjs.com/ it says in small writting "
Express is a project of the Node.js Foundation".

Doug Wilson was at a Node Interactive talking about it etc.

Day to day there has been little linkage/interaction but before we delete what might be the only doc, we should probably work out what the situation is (likely through @MylesBorins with discussion with the Board).  @jasnell may also be able to chime in with some more context.

jasnell:
Express is still technically a foundation project but there has been little to no interaction with the foundation due to there simply not needing to be any. The project is functioning well with it's own TC.

Trott:
> -1. I see no reason to remove this at this time.

It needs to be removed because  it is incorrect and misleading.

* It trains us to  ignore our documented processes.

* It makes accurate information harder to find.

* It fosters misconceptions for newcomers reading the docs for the first time. It encourages them to engage in fruitless processes and then makes them feel lied to.

Trott:
TL;DR: This change doesn't say "express is no longer a Node.js Foundation project" (whatever that statement means). It just says "There's no such thing as a Top-Level Project or Incubator Program." Because there isn't. 

> Express was brought under the auspices of the Node.js foundation. Look at "Applications/Express.md". 

@mhdawson I'm looking. I still don't know what "brought under the auspices of the Node.js Foundation" means in this context. 

> Look here https://expressjs.com/ it says in small writting "Express is a project of the Node.js Foundation".

Yup. Still don't understand what that means.

As far as I can tell, Express was admitted to the "Incubator program" in February 2016. The Incubator program, as far as I can tell, just means that they get some kind of mentorship/help with governance processes.

Seems like that process has ended. They've got their governance together, things are running well enough. If they want more assistance, they can certainly ask.

The Incubator Program does not exist anymore, to the extent that it ever actually existed.

This isn't what we do and the documentation here makes it sound like it's a Big Deal. It's not. It's not even an afterthought. It's an experiment that didn't pan out and something we don't do anymore. Let's get rid of it.

As far as what it means that express is "under the Node.js Foundation", I don't think anyone knows what that actually means. It *seems* to mean: We provide symbolic-only support and in the event that something goes sideways with express, the Node.js Foundation is committed to stepping in and trying to help right things. None of that has anything to do with the material being removed in this pull request IMO.

Trott:
Seems like perhaps the "express is under the foundation" thing is summarized here? https://github.com/nodejs/TSC/pull/39#issuecomment-178242355

Seems like all the stuff in there that the Foundation might do (like managing the domain) is relegated in the doc to IBM. Which is totally fine.

It does seem like any of the things suggested in that comment as significant aspects of being "in the foundation" are things the board would need to approve. I'm trying to think about where the best place to document such a decision would be. It's not here. Website? Main repo? Somewhere else? 

Do we even know if this is something that the express folks still want/value?

jasnell:
> The Incubator program does not exist anymore, to the extent that it ever actually existed.

Yes, it did. @mikeal and I filled that role for express and helped with a number of items, such as helping express get it's github organization upgraded too a free tier that would allow private repositories to be created. Since that time, the project has evolved past the need for any further guidance from the TSC and stands alone as it's own wholly separate and independent top level project within the Foundation.

The documentation around top level projects can certainly be improved, and the process refined, but removing it entirely does not seem appropriate to me. It may be that the TSC wishes to recuse itself from any responsibility around admitting new top level projects like express under the foundation, but that should be handled differently if that is the case.

jasnell:
libuv is another example. It *is also* a top level project under the Node.js Foundation that operates independently of the Node.js TSC. It has it's own governance and it's own set of contributors. The TSC has not needed to provide any mentorship for the project as far as I know. The TSC may wish to explicitly state that it has no direct involvement in libuv, but that doesn't make it any less of a top level project.

Trott:
> > The Incubator program does not exist anymore, to the extent that it ever actually existed.
>
> Yes, it did.

I didn't say it didn't. You seem to have misread the text your quoting and responding to.

> @mikeal and I filled that role for express and helped with a number of items, such as helping express get it's github organization upgraded too a free tier that would allow private repositories to be created. Since that time, the project has evolved past the need for any further guidance from the TSC and stands alone as it's own wholly separate and independent top level project within the Foundation.

Thanks for the additional information and context. Yes, that's a thing that happened. Express was admitted to the incubator project, some stuff happened, and then that was basically that. You're calling it a "top-level project" but it is not a top-level project _as that term is defined in these docs_. 

* The doc says that the TSC oversees all top-level projects. We do not oversee express. The doc explicitly calls out technical oversight. We do no such thing.

* The doc says that top-level projects have a representative on the TSC. express does not.

* The doc says that the TSC is responsible for admitting top-level projects. The TSC never admitted express as a top-level project. They were admitted to the incubator program. Clearly there was an intention to eventually graduate them from that to TLP status, but that never happened, mostly because the whole TLP thing isn't really necessary, at least not from a Node.js governance perspective and seemingly not from the express side either, so everyone forgot about it. Maybe the incubator program is useful (but should probably be called _governance mentorship_ or something like that). And maybe having legal and IP and whatnot under the board is useful. But those things do not make something a TLP. At least not according to our docs.

>  libuv is another example.

No, it is not, for all the reasons listed for express above.

> It is also a top level project under the Node.js Foundation that operates independently of the Node.js TSC. 

That description explicitly contradicts the description of TLPs in this doc.

Trott:
I'm going to close this and open something smaller/milder in the hopes of gaining some consensus. My first concern is that stuff that is absolutely positively 100% false be removed or corrected, so I'll try to limit myself to those changes.

mhdawson:
@Trott I'm all for fixing up things that are out of date, look forward to you follow on PRs.

PR #469, opened 2018-01-12
Moderation access

mhdawson:


Trott:
Can you please run this (and all doc contributions) through a spelling/grammar checker?

mhdawson:
One note is that we are still moving some repos over to nodejs-private, and we should probably wait until that is done to land this (if we reach consensus).  I don't think that moving them over should take more than a few days.

GitHub-Org-Management-Policy.md
```diff
@@ -16,8 +16,25 @@ Committee members have write permissions to the Node.js admin repository.
...
 
-TSC members and the Chair of the Community Committee are the only individuals granted
-Owner permissions within the Node.js GitHub Organization.
+As with other resources within the community, Org ownership will be granted in order
+to optimize for the conflicting requirements of:
+
+* Limiting access to reduce risk
+* Enabling individuals to move community work forward without undue delay
+
+When possible we should use automation and tools to reduce the breadth of access
```
Trott: add comma after `possible`

mhdawson:
@Trott  do you have a good suggestion for one that you use when generating md files ?

GitHub-Org-Management-Policy.md
Trott: Probably don't want to use `we` here or on line 27.

GitHub-Org-Management-Policy.md
```diff
@@ -16,8 +16,25 @@ Committee members have write permissions to the Node.js admin repository.
...
+that needs to be provide in order to enable individuals to move community work foward.
```
Trott: typo: `be provide` -> `be provided`

Trott: `forward` is misspelled

GitHub-Org-Management-Policy.md
```diff
@@ -16,8 +16,25 @@ Committee members have write permissions to the Node.js admin repository.
...
+As these tools are developed and put in place we will reduce the groups to which
+org ownership is granted.
+
+At the current time the following groups are granted access in order for them
```
Trott: add a comma after `time`

GitHub-Org-Management-Policy.md
```diff
@@ -16,8 +16,25 @@ Committee members have write permissions to the Node.js admin repository.
...
+to facilitate community work.
+
+* TSC members
+* Chair of the Community Comittee
```
Trott: `Committee` is misspelled

GitHub-Org-Management-Policy.md
```diff
@@ -16,8 +16,25 @@ Committee members have write permissions to the Node.js admin repository.
...
+* Moderation team members. - It is expected that moderation team members will limit
+their use of the access granted to that requied to carry out moderation across the
```
Trott: `required` is misspelled

Trott:
> @Trott do you have a good suggestion for one that you use when generating md files ?

I usually use the free version of Grammarly, although I don't know how well it does with raw markdown. I have a Markdown editor, so I copy/paste the rendered doc into Grammarly rather than the raw markdown. But I'd guess it's fine, especially for a doc like this. It can generate a lot of false positives, but I think that's pretty much true of all the grammar tools.

mhdawson:
Pushed commit addressing initial comments.

GitHub-Org-Management-Policy.md
```diff
@@ -16,8 +16,25 @@ Committee members have write permissions to the Node.js admin repository.
 
 ### Owners
 
-TSC members and the Chair of the Community Committee are the only individuals granted
-Owner permissions within the Node.js GitHub Organization.
+As with other resources within the community, Org ownership will be
+granted in order to optimize for the conflicting requirements of:
```
Trott: We don't use "Org ownership" in the doc. That phrase is pretty colloquial. Stick to the terminology elsewhere: Owner permissions.

"As with other resources within the community..." is filler. Take it out.

So, all told, maybe something like this?:

> Whether to grant of Owner permissions is determined by optimizing for the following conflicting requirements:

(As an alternative, I'd also be all for taking out all of this explanatory material and just identifying the people that have Owner permissions.)

GitHub-Org-Management-Policy.md
```diff
@@ -16,8 +16,25 @@ Committee members have write permissions to the Node.js admin repository.
...
+
+* Limiting access to reduce risk
+* Enabling individuals to move community work forward without undue delay
+
+When possible, automation and tools should be used to reduce the breadth of
+access that needs to be provided in order to enable individuals to move
+community work forward.  As these tools are developed and put in place we will
```
Trott: How about this instead?:

> As these tools are developed and put in place, the groups to which Owner permissions are granted will be reduced.

GitHub-Org-Management-Policy.md
```diff
@@ -16,8 +16,25 @@ Committee members have write permissions to the Node.js admin repository.
...
+reduce the groups to which org ownership is granted.
+
+At the current time, the following groups are granted access in order for them
```
Trott: Nit: "At the current time" is superfluous. Take it out.

Trott: Also consider removing "in order for them to facilitate community work.". The sentence should end with a colon as well. So all told, maybe this?:

> The following groups are granted Ownership permissions:

GitHub-Org-Management-Policy.md
```diff
@@ -16,8 +16,25 @@ Committee members have write permissions to the Node.js admin repository.
...
+to facilitate community work.
+
+* TSC members
+* Chair of the Community Committee
+* Moderation team members. - It is expected that moderation team members
```
Trott: Nit: remove the dash/hyphen/whatever.

Trott: Nit: Get rid of "It is expact that".

Trott: Generally, I think we've been styling it captialized in the docs: `Moderation Team`

Trott:
The TL;DR of this seems to be "Moderation Team gets Owner permissions until there's some other way for them to do their work, which will hopefully be a bot at some point in the not-too-distant future." 

+1 from me!

jasnell:
Given the objections that @rvagg has raised on this, I'd like for this to to on the tsc-agenda for discussion before it lands. I'm not -1'ing the actual change, I just want it to be discussed on the call.

mhdawson:
Was discussed in last TSC meeting, no objections to moving forward.  James said we could dismiss his objection.

jasnell:
Also from the TSC meeting: Everyone with owner access needs to have 2FA enabled on their github account. That needs to be confirmed before assigning the owner permission. Also, efforts to pursue automation options need to move forward so that assigning owner permissions to individuals outside the TSC and CommComm chairs no longer becomes necessary. The rationale there is strictly limited to general security concerns -- that is, the higher the number of users with owner permissions, the higher the risk of something bad happening should one of those accounts be compromised. So granting owner permissions to the moderation team should be considered a temporary action *until* an automated solution can be developed.

mhdawson:
@Trott took the action to validate 2FA before adding owner access.

jasnell:
> took the action to validate 2FA before adding owner access

Yep... for existing mod team members... it needs to be confirmed for *new* members before assigning permissions.

mhdawson:
Landed as 9cf274c532ca22bf91e43d6d87af1e6aac59fd0f

PR #473, opened 2018-01-18
retire mod team from strategic initiatives

Trott:
The Moderation Team is spun up and operational. It is now an ongoing
day-to-day activity rather than a strategic initiative. It will still
need to report weekly activity to TSC and CommComm, but it is no longer
necessary to include it in the list of Strategic Initiatives.

By the way, this might be a good opportunity for others to look for items that may no longer need to be treated as strategic.

### DIFF
--- a/Strategic-Initiatives.md
+++ b/Strategic-Initiatives.md
@@ -22,7 +22,6 @@ and have the support needed.
 | Core Promise APIs | James Snell                     |                                                                  |
 | TSC Governance    | Myles Borins                    | https://github.com/nodejs/TSC/issues/383                         |
 | New Streams APIs  | James Snell + Jeremiah Senkpiel | https://github.com/nodejs/node/pull/16414                        |
-| Moderation        | Rich Trott                      |                                                                  |
 | Async Hooks       | Trevor Norris                   |                                                                  |
 | V8 Currency       | Michaël Zasso                   |                                                                  |
 


maclover7:
Should this be moved to the "completed" section?

Trott:
> Should this be moved to the "completed" section?

Could be. I'm not sure. Happy to do it that way if that's what others think. I guess I lean towards this: I don't like the idea of thinking that something like moderation is "completed" or "done". "Operational and no longer strategic"? Yup. Completed? Not so much. But maybe I'm overthinking.

jasnell:
hmm... not sure we should. There is a requirement in the Moderation policy that the moderation team provide a regular report of activities to both the TSC and CommComm. That requirement has been served by including the mod team in our strategic initiatives run-through each week and it's been quite valuable, I think.

gibfahn:
>>It will still need to report weekly activity to TSC and CommComm, but it is no longer necessary to include it in the list of Strategic Initiatives.

>There is a requirement in the Moderation policy that the moderation team provide a regular report of activities to both the TSC and CommComm. That requirement has been served by including the mod team in our strategic initiatives run-through each week and it's been quite valuable, I think.

@Trott if it's not in the strategic initiatives, will it still show up in the agenda?

Trott:
> @Trott if it's not in the strategic initiatives, will it still show up in the agenda?

Nope. I don't think it needs to. The Mod Team charge/charter/whatever says it needs to report monthly. It's up to them to figure out how the report should happen and when. (Currently, we just do it weekly during the TSC/CommComm meetings, but the Mod Team is free to change that as they see fit within the parameters in the charter.) It's not TSC's responsibility to put it on the agenda, as the report doesn't have to happen during meetings. (It could be emailed monthly, for example.)

Trott:
While I'm at it: Meetings should be the last-resort method of communicating anything anyway. GitHub issues, discussion boards, and private email are all vastly better. For one thing they guarantee that everyone gets the same information at the same time. (Meetings rarely have more than maybe 70% attendance due to time zones and other scheduling issues.) For another, they all leave at least some form of paper trail. Meeting minutes are a pale substitute IMO.

TL;DR: Anything that can be removed from the meeting agenda and discussed elsewhere should be.

gibfahn:
To be clear, I think saying it in the meeting is unnecessary, I just think it should be in the issue and in the meeting notes. https://github.com/nodejs/TSC/issues/478#issuecomment-362001185 is fine, but there isn't a reminder to do it.

PR #479, opened 2018-01-31
Node.js OpenSSL Strategy

rvagg:
Here's my proposal for Node.js' OpenSSL Strategy. This document includes the following:

 * Background, including how things work for Node versions 4 to 9
 * How Node.js binaries ship and are used in practice in relation to OpenSSL
 * OpenSSL release lines and their LTS policy (or lack thereof)
 * OpenSSL 1.1.0
 * OpenSSL 1.1.1
 * OpenSSL FIPS Module and Node.js' use of it
 * **OpenSSL strategy for Node.js 10**
 * OpenSSL forks: LibreSSL and BoringSSL

---------------------

***Previous OP:***

~~This is very WIP but I thought I'd better start getting this up rather than let it potentially go stale on my computer if I can't get back to it soon. Missing key pieces, particularly the Node 10 section which I'm going to use to make a recommendation. There's some good context in here thought which might behelpful.~~

~~I'll ping the TSC and crypto team when it's ready to be read and considered in full.~~

~~Also, I'm not sure whether this is the best place for this. The other two options I considered were nodejs/Release and nodejs/node/deps/openssl/doc. Happy to hear thoughts on this.~~


OpenSSL-Strategy.md
```diff
@@ -0,0 +1,50 @@
+# Node.js OpenSSL Strategy
+
+The Node.js source tree contains a copy of OpenSSL. By default, Node.js compiles its own copy of OpenSSL and statically links this library into its binary. The `configure` tool allows for dynamic linking against a shared OpenSSL library. Whether this linking is possible is left for the compiler to determine at build and linking time.
+
+## Node.js versions 4 to 9
+
+Identical copies of the latest OpenSSL 1.0.2 version are included in Node.js releases from versions 4 to 9. Aside from some manual configuration that is required in order to support GYP builds (instead of the Python-based Configure script that OpenSSL ships) as described in [deps/openssl/doc/UPGRADING.md](https://github.com/nodejs/node/blob/master/deps/openssl/doc/UPGRADING.md), there are 4 floating patches that are applied on top of the plain OpenSSL 1.0.2 source:
```
bnoordhuis: s/Python-based/Perl-based/

OpenSSL-Strategy.md
```diff
@@ -0,0 +1,50 @@
...
+* **NodeSource Linux Distributions**: Binaries shipped by this heavily used source do not follow Linux distribution policy and use the default `configure` configuration for OpenSSL and therefore compile and statically link in the OpenSSL source as it ships with Node.js.
+* **Homebrew**: The [formula](https://github.com/Homebrew/homebrew-core/blob/master/Formula/node.rb) for Node.js as distributed on macOS by the popular `brew` command, by default, will compile with using the default `configure` configuration. It is possible to override this using a `with-openssl` option which will compile against the version of OpenSSL that was most recently installed with `brew` but this is not believed to be in common use.
```
danbev: Nit: perhaps remote `with` in `command, by default, will compile with using the default 'configure'

rvagg:
@nodejs/tsc @nodejs/security-wg @nodejs/crypto here's my OpenSSL Policy proposal. I'd like to get this in discussion ASAP as it has a big impact on Node 10 thanks to the expiry of 1.0.2 support 1/2 way through the Node 10 lifecycle. It's not as straightforward as just upgrading, as you'll read. My proposal for 10 is toward the bottom but the context is all above it, so don't skip _too much_.

Relevant to this discussion is that OpenSSL 1.1.1-pre1 has already been released and there is [some work underway](https://github.com/nodejs/node/issues/18770) to test and make Node compatible. TLS 1.3 hasn't been finalized yet and we don't have much of an idea for a release date for 1.1.1. Very unlikely to be prior to Node 10 launch although we may have a clearer idea as we get closer. I'll push on the OpenSSL team as we get closer and just maybe we could shuffle our release timing to get straight to 1.1.1 and skip this 1.1.0 awkwardness. But for now we just have to assume it's not going to be a possibility and plan accordingly.

OpenSSL-Strategy.md
```diff
@@ -0,0 +1,88 @@
...
+
+Node.js compiled against the OpenSSL FIPS Object Module does not make Node.js itself FIPS validated. Instructions for compiling Node.js against the Object Module along with links to further documentation are detailed in [BUILDING.md](https://github.com/nodejs/node/blob/master/BUILDING.md#building-nodejs-with-fips-compliant-openssl).
+
+The OpenSSL FIPS Object Module is compatible with OpenSSL 1.0.2 and Node.js has been able to build with this module since 2015, prior to Node.js 4. It requires some modification of the Node.js internals (see `git grep FIPS -- lib/ src/`) for this to work properly.
+
+Development and validation of a FIPS software component is time consuming and expensive. The OpenSSL team have yet to commit to a timeframe for development of the next generation of its FIPS Object Module, however they have stated that it is their next priority ["after 1.1.1"](https://www.openssl.org/policies/roadmap.html). Therefore, any user requiring FIPS validated
```
bnoordhuis: Seems like the end is missing.

shigeki:
Here is the current timeline of forthcoming 1.1.1.

https://www.openssl.org/policies/releasestrat.html

```
13th February 2018, alpha release 1 (pre1)
27th February 2018, alpha release 2 (pre2)
13th March 2018, beta release 1 (pre3)
OpenSSL_1_1_1-stable created (feature freeze)
master becomes basis for 1.1.2 or 1.2.0 (TBD)
27th March 2018, beta release 2 (pre4)
10th April 2018, beta release 3 (pre5)
24th April 2018, beta release 4 (pre6)
1st May 2018, release readiness check (new release cycles added if required, first possible final release date: 8th May 2018)
```
And EOSL of 1.1.0 was changed as
```
Version 1.1.0 will be supported until one year after the release of 1.1.1
```

TLS 1.3 is one of the main features to be expected to use 1.1.1 and is enabled by default. 
It is now in IETF last call until 2018-03-02 and IESG discussion is to be held in 2018-03-08. https://datatracker.ietf.org/doc/draft-ietf-tls-tls13/  
I think we can see if TLS1.3 can be finalized before 1.1.1 in March. 

The date of FIPS support in 1.1.x is not yet to be known.
https://www.openssl.org/blog/blog/2017/08/17/fips/

rvagg:
nit fixed

@shigeki thanks for the additional info, I hadn't seen that they outlined a 1.1.1 release timeframe and the TLS 1.3 timeframe clarity is helpful. I updated the doc:

> OpenSSL 1.1.1 pre-releases are already available as of the time of writing. According to the [OpenSSL release strategy](https://www.openssl.org/policies/releasestrat.html), the 1st of May is the target for "release readiness", with the first possible release date being the 8th of May if the codebase is considered ready for release.
>
> In addition, the TLS 1.3 specification is in "Last Call" state, due to end on the 2nd of March. This means that it is likely that we will receive a finalized TLS 1.3 specification during March, which would make a May OpenSSL 1.1.1 release feasable.
>
> TLS 1.3 specification finalization and code readiness are the two reasons that OpenSSL 1.1.1 release in early May is not a certainty. However, current signs point to that probability.

@bnoordhuis my concern is with distros that will continue to support 1.0.2 like RedHat beyond its EOL and may still want to bundle Node with 1.0.2, hence the note about continuing support. However, I've changed it now to say:

> * Remain backward-compatible with OpenSSL 1.0.2 via dynamic linking for the lifetime of Node.js 10. This will cease being an _official_ policy at the OpenSSL 1.0.2 EOL date which will occur 4 months prior to Node 10 entering Maintenance LTS. However, this support will be maintained as long as it does not cause security problems and there are contributors available to maintain such support. The lack of a passing test suite with Node.js 10 compiled against OpenSSL 1.0.2 past the EOL date will not hold up further releases of Node 10.

How does that sound? Ideally we'd continue to support 1.0.2 but we're not going to hold ourselves to that commitment beyond EOL if it's too hard for whatever reason.

bnoordhuis:
Sounds good!

mcollina:
:+1: on that wording.

OpenSSL-Strategy.md
```diff
@@ -0,0 +1,94 @@
...
+
+The [OpenSSL Roadmap](https://www.openssl.org/policies/roadmap.html) does not yet outline a new LTS version, however the OpenSSL team have variously made statements (e.g. [here](https://github.com/nodejs/node/issues/4270#issuecomment-355151355)) that a new LTS release line will be designated at some point in 2018.
+
+As the 1.0.2 support timeframe does not extend for the entirety of what should be the support timeframe for Node.js version 8 (2020-04), the version 8 EOL date has been [contracted](https://github.com/nodejs/release#release-schedule) to coincide with the OpenSSL 1.0.2 EOL date at the end of 2019.
+
+### OpenSSL 1.1.0
+
+OpenSSL 1.1.0 represents a fairly major rework of the codebase, at least in comparison to its slow evolution until this point. The external API has some major departures from 1.0.2. However, support for compiling against an external OpenSSL 1.1.0 library (dynamically linked) was [added](https://github.com/nodejs/node/pull/16130) to Node.js' master branch in late 2017.
+
+Even though OpenSSL 1.1.0 is only supported until August 2018, the API shift is an important stage in Node.js' adaptation.
```
shigeki: The support of 1.1.0 is extended to one year after 1.1.1 release as in https://www.openssl.org/policies/releasestrat.html. So Aug. 2018 is no longer support dead line.

OpenSSL-Strategy.md
```diff
@@ -0,0 +1,94 @@
...
+* **NodeSource Linux Distributions**: Binaries shipped by this heavily used source do not follow Linux distribution policy and use the default `configure` configuration for OpenSSL and therefore compile and statically link in the OpenSSL source as it ships with Node.js.
+* **Homebrew**: The [formula](https://github.com/Homebrew/homebrew-core/blob/master/Formula/node.rb) for Node.js as distributed on macOS by the popular `brew` command, by default, will compile using the default `configure` configuration. It is possible to override this using a `with-openssl` option which will compile against the version of OpenSSL that was most recently installed with `brew` but this is not believed to be in common use.
+
+## OpenSSL release lines and LTS
+
+As of writing, there are two officially supported versions of OpenSSL as per the [OpenSSL Release Strategy](https://www.openssl.org/policies/releasestrat.html)
+
+* Version 1.0.2: designated Long-term Support (LTS) which will be supported until 2019-12-31
+* Version 1.1.0: supported until 2018-08-31
```
shigeki: it is changed as described below.

OpenSSL-Strategy.md
```diff
@@ -0,0 +1,94 @@
...
+Development and validation of a FIPS software component is time consuming and expensive. The OpenSSL team have yet to commit to a timeframe for development of the next generation of its FIPS Object Module, however they have stated that it is their next priority ["after 1.1.1"](https://www.openssl.org/policies/roadmap.html). Therefore, any user requiring FIPS validated OpenSSL must use OpenSSL 1.0.2 until a new module is made available that supports a newer version of OpenSSL.
+
+## Node.js 10
+
+OpenSSL LTS support timing, the lack of OpenSSL LTS planning and the lack of a clear timeframe for a new FIPS module complicates Node.js 10.
+
+As of the time of writing, the strategy for OpenSSL with Node.js 10 is:
+
+ * OpenSSL 1.1.0 to initially be the assumed default compile target.
+ * Bundle a copy of OpenSSL 1.1.0 in the source tree, along with any floating patches still required for improved Windows support and test-runner speed-ups.
```
shigeki: I think we should add the following a new build requirement especially for Windows.

* A new build requirement of NASM is needed on Windows as in 
https://github.com/openssl/openssl/blob/89314d9a22ad0b0ce87e188c486532926aff01ea/NOTES.WIN#L21-L24 .
Microsoft assembler, `masm`, is no longer supported in building Node.js with OpenSSL-1.1.x.

OpenSSL-Strategy.md
```diff
@@ -0,0 +1,94 @@
+# Node.js OpenSSL Strategy
+
+The Node.js source tree contains a copy of OpenSSL. By default, Node.js compiles its own copy of OpenSSL and statically links this library into its binary. The `configure` tool allows for dynamic linking against a shared OpenSSL library. Using a combination of the `--shared-openssl`, `--shared-openssl-includes`, `--shared-openssl-libname` and `--shared-openssl-libpath` arguments to `configure`, compilation of Node.js will skip over its bundled copy of the OpenSSL source and link to dynamic library version of OpenSSL, or a suitably compatible OpenSSL equivalent. Whether this linking is possible is left for the compiler to determine at build and linking time.
+
```
shigeki: We have a build requirements of assembler version as described in https://github.com/nodejs/node/blob/3a191229418dcc0e21956847993b1702c88a923b/deps/openssl/openssl.gypi#L1038-L1039 and we are using masm (Microsoft assembler) is used building Node.js despite of OpenSSL recommendation in historical reason, which is need to be deprecated in building OpenSSL-1.1.x.
Theses are can be written as below.
```
The following assembler versions are needed to support in AVX and AVX2 features of OpenSSL-1.0.2 in building  Node.js.  'llvm_version>="3.3" or xcode_version>="5.0" or gas_version>="2.23"
In windows, Microsoft assembler, `masm` can be supported in building Node.js with OpenSSL-1.0.2 in despite of OpenSSL recommendation.
```

OpenSSL-Strategy.md
shigeki: A newer gas or llvm version would be required for use of AVX512 features but it has slightly different version requirements between 1.1.0 and 1.1.1.
 I think it is one the issues to be resolved before include OpenSSL-1.1.x build binding with Node

OpenSSL-Strategy.md
```diff
@@ -0,0 +1,94 @@
...
+Even though OpenSSL 1.1.0 is only supported until August 2018, the API shift is an important stage in Node.js' adaptation.
+
+### OpenSSL 1.1.1
+
+As per the OpenSSL Roadmap, the focus of current development on the 1.1.1 release line is TLS 1.3. Unfortunately, the OpenSSL team is not currently prepared to designate 1.1.1 the next TLS line but have made a commitment that it will be both API and ABI compatible with OpenSSL 1.1.0 (see [here](https://www.openssl.org/blog/blog/2017/05/04/tlsv1.3/) and [here](https://github.com/openssl/openssl/issues/5120#issuecomment-359212121)).
+
+OpenSSL 1.1.1 pre-releases are already available as of the time of writing. According to the [OpenSSL release strategy](https://www.openssl.org/policies/releasestrat.html), the 1st of May is the target for "release readiness", with the first possible release date being the 8th of May if the codebase is considered ready for release.
+
+In addition, the TLS 1.3 specification is in "Last Call" state, due to end on the 2nd of March. This means that it is likely that we will receive a finalized TLS 1.3 specification during March, which would make a May OpenSSL 1.1.1 release feasable.
```
Trott: Micro-nit: Typo: `feasable` -> `feasible`

OpenSSL-Strategy.md
```diff
@@ -0,0 +1,94 @@
+# Node.js OpenSSL Strategy
+
+The Node.js source tree contains a copy of OpenSSL. By default, Node.js compiles its own copy of OpenSSL and statically links this library into its binary. The `configure` tool allows for dynamic linking against a shared OpenSSL library. Using a combination of the `--shared-openssl`, `--shared-openssl-includes`, `--shared-openssl-libname` and `--shared-openssl-libpath` arguments to `configure`, compilation of Node.js will skip over its bundled copy of the OpenSSL source and link to dynamic library version of OpenSSL, or a suitably compatible OpenSSL equivalent. Whether this linking is possible is left for the compiler to determine at build and linking time.
```
gibfahn: Nits/suggestions:

```diff
-By default, Node.js compiles its own copy of OpenSSL and statically links this library into its binary.
+By default, the Node.js build process compiles its own copy of OpenSSL and statically links this library into the built `node` binary.
```

```diff
-link to dynamic library version of OpenSSL
+link to a dynamic library version of OpenSSL
```

```diff
-at build and link time.
+at build and linking time.
```

Also it would be good to format to 80 columns, not least so leaving PR comments is easier.

OpenSSL-Strategy.md
```diff
@@ -0,0 +1,94 @@
...
+
+## Node.js versions 4 to 9
```
gibfahn: Nit: maybe `4.x to 9.x` for clarity

OpenSSL-Strategy.md
```diff
@@ -0,0 +1,94 @@
...
+* **Linux Distributions**: Binaries shipped with Debian, Ubuntu, CentOS, RHEL, Fedora and the majority of other Linux distributions follow a standard policy of separating dependencies and dynamically linking wherever possible. Therefore, the Node.js packages on these systems are linked via the packaging dependency mechanisms to OpenSSL 1.0.2 packages and the `node` binaries that ship by default on these platforms use the shared OpenSSL 1.0.2 library installed on that by those packages. Therefore, the floating patches do not apply and it is possible that a different version of OpenSSL 1.0.2 is in use by Node.js than the version that was shipped in the source tree.
+* **NodeSource Linux Distributions**: Binaries shipped by this heavily used source do not follow Linux distribution policy and use the default `configure` configuration for OpenSSL and therefore compile and statically link in the OpenSSL source as it ships with Node.js.
```
gibfahn: Would be good to add a link, maybe to https://github.com/nodesource/distributions ?

OpenSSL-Strategy.md
```diff
@@ -0,0 +1,94 @@
...
+* [2eb170874](https://github.com/nodejs/node/commit/2eb170874aa5e84e71b62caab7ac9792fd59c10f): `openssl: fix keypress requirement in apps on win32`. A fix for the `openssl` client application, used by the Node.js test suite, to properly accept stdin without requiring a physical keyboard. As described in <http://openssl.6102.n7.nabble.com/PATCH-s-client-Fix-keypress-requirement-with-redirected-input-on-Windows-td46787.html>.
+* [664a65969](https://github.com/nodejs/node/commit/664a6596960655e214fef25e74d3285097703e95): `deps: add -no_rand_screen to openssl s_client`. Adds a `-no_rand_screen` command line option to the `openssl` client application, used by the Node.js test suite which skips invocation of the `RAND_screen()` call on Windows in order to speed up some of the Node.js TLS tests. Use of this can be found in many of the `test/*/test-{tls,https}-*` test files.
```
gibfahn: It would be really useful to have a `maintaining-openssl.md` (similar to the existing [`maintaining-v8.md'](https://github.com/nodejs/node/blob/master/doc/guides/maintaining-V8.md) and [`maintaining-npm.md`](https://github.com/nodejs/node/blob/master/doc/guides/maintaining-npm.md) guides, that includes a link to this list so we don't forget to update it.

Not really a job for this PR, maybe a request for @shigeki.

OpenSSL-Strategy.md
```diff
@@ -0,0 +1,94 @@
...
+ * **No support for FIPS** unless a new FIPS module becomes available during the Node.js 10 lifetime and is compatible without requiring breaking changes.
+
+ABI and API compatibility cannot be guaranteed in a switch from 1.1.0 to 1.1.1 although, as previously mentioned, the OpenSSL team have signaled their intention for this to be the case. The Node.js team should work with the OpenSSL team to ensure this is the case and smooth the upgrade path.
+
+The lack of FIPS support is unfortunate, however, unless a new FIPS module takes an inordinate amount of time, Node.js users requiring FIPS support should be able to use Node.js 8 and switch to a future Node.js version that supports the new FIPS module (ideally Node.js 12).
+
+This strategy must be communicated to users of Node.js 10 early and often. There is potential for instability and a change in default OpenSSL version is unprecedented and therefore unexpected. The potential for breaking API and/or ABI may also cause disruption, potentially requiring an increment of `NODE_MODULE_VERISION`, which will also be unprecedented within a single release line. It is important that users be aware of this possibility.
```
gibfahn: `NODE_MODULE_VERISION` -> `NODE_MODULE_VERSION`

OpenSSL-Strategy.md
richardlau: There's https://github.com/nodejs/node/wiki/OpenSSL-upgrade-process, although it hasn't been updated since 2016.

OpenSSL-Strategy.md
mhdawson: So from that doc it is 8 May 2019 at the latest.

mhdawson:
Thinking about it a bit more, @rvagg what would your take be on waiting 1 month on Node 10.x (target end of May instead of April).  It might be worth it to avoid the potential of having to update the NODE_MODULE_VERSION

MylesBorins:
@mhdawson I'm a big +1 in delaying the release if we need to for consistency (we have prior art of this with 8.x). Seems like something to discuss in the meeting today

jasnell:
I am -1 on delaying 10.0.0 at this point as I'm not certain that FIPS support is something we should be blocking on at all. We need to assess the likelihood of ABI/API breaks in 1.1.1 before suggesting delaying 10.0.0

OpenSSL-Strategy.md
yhwang: Nit: In Windows, we use `vcbuild.bat` and the `configure_flags` parameter is used to pass the options that you mentioned above for `configure` tool. Not sure should we also cover the usage in Windows or not.

jasnell:
Specifically, I'm +1 on the plan on shipping 10.0.0 with 1.1.0 on the regular schedule (current release target date April 24th) and moving up to 1.1.1 as it becomes available.

gibfahn:
>Specifically, I'm +1 on the plan on shipping 10.0.0 with 1.1.0 on the regular schedule (current release target date April 24th) and moving up to 1.1.1 as it becomes available.

Seems indicated to me given the assurance from the OpenSSL team that it'll be API and ABI compatible.

shigeki:
> Seems indicated to me given the assurance from the OpenSSL team that 1.1.1 will be API and ABI compatible.

We have it in https://github.com/openssl/openssl/issues/5120#issuecomment-368064230.

I have the following two concerns about the plan of Node-v10 to start OpenSSL-1.1.0 then move to 1.1.1.

1. Assembler version requirements

They are different between 1.1.0 and 1.1.1 as below.

OpenSSL-1.1.0 minimum assembler version requirements
https://github.com/openssl/openssl/blob/OpenSSL_1_1_0-stable/doc/crypto/OPENSSL_ia32cap.pod

Extension   | GNU as | nasm   | llvm
------------|--------|--------|--------
AVX         | 2.19   | 2.09   | 3.0
AVX2        | 2.22   | 2.10   | 3.1
AVX512      | 2.25   | 2.11.8 | 3.6

OpenSSL-1.1.1 minimum assembler version requirements
https://github.com/openssl/openssl/blob/master/doc/man3/OPENSSL_ia32cap.pod

Extension   | GNU as | nasm   | llvm
------------|--------|--------|--------
AVX         | 2.19   | 2.09   | 3.0
AVX2        | 2.22   | 2.10   | 3.1
ADCX/ADOX   | 2.23   | 2.10   | 3.3
AVX512      | 2.25   | 2.11.8 | see NOTES
AVX512IFMA  | 2.26   | 2.11.8 | see NOTES
VAES        | n/a    | n/a    |

We are pre-generating assembler files while I am doing  with upgrading OpenSSL. If we cannot change build requirements after Node-10 release,  it cannot use of AVX512IFMA feature. Otherwise, we needs to pre-generate several kind of assembler files depending on assembler versions. But it leads to be more complex in the future upgrading procedures of OpenSSL.

2. New features after moving to OpenSSL-1.1.1

Shared library users of Node-v10 will continue to use OpenSSL-1.1.0 after we upgrade to 1.1.1 so that we have to maintain two versions of OpenSSL until EOLS of 1.1.0, one year after 1.1.1.
If we add a new feature specific to 1.1.1, the API should check openssl version and be noted in the API doc. 

For the API simplicity, I think it is better to support one release version of OpenSSL during LTS.

shigeki:
My PR to upgrade OpenSSL-1.1.0 is coming in the next week after the forth-coming security release of OpenSSL-1.1.0h. 

Here is my WIP branch at  https://github.com/shigeki/node/tree/upgrade_openssl110h_pre
and CI results are fine except Jenkins errors in https://ci.nodejs.org/job/node-test-commit/17127/.

rvagg:
TLS 1.3 got finalised this week, so 1.1.1 has another box ticked https://www.ietf.org/mail-archive/web/ietf-announce/current/msg17592.html

fhinkel:
Is this ready to land?

mhdawson:
@shigeki you have a blocking objection, are there still updates needed to the doc?

rvagg:
Most excellent news @nodejs/tsc: https://www.openssl.org/blog/blog/2018/05/18/new-lts/

> Our current LTS release is 1.0.2, and it will be supported until the end of 2019. During that last year it will only receive security fixes. Although we are extended 1.1.0 support, we explicitly decided not to do it again, for either release.
>
> Our next LTS release will be 1.1.1 which is currently in beta. As long as the release is out before the end of 2018, there is more than a year to migrate. (We’re confident it will be out before then, of course.) We encourage everyone to start porting to the OpenSSL master branch.

There's some notes about FIPS too:

> we’re committed to a new FIPS module, it will be based on the current codebase, and we think we can get it done fairly quickly.

As I said in the last TSC meeting, 1.1.1 is pushing further than originally hoped but we're already up to beta 7 as of yesterday and I think things are getting ironed out and ready for release. Beta 8 is due on the 19th of June and they so far haven't planned anything beyond that. There's a release criteria for 1.1.1 at the bottom of https://www.openssl.org/policies/releasestrat.html which basically amounts to not having outstanding major issues tagged for the branch. So cross fingers and we may see a 1.1.1 within the next month.

Of course there is a risk of this dragging on closer to October when we hit LTS. After this point an upgrade is going to be really annoying. I think the current indications are that we'll make it long before that but we need to be prepared to think through the implications of that upgrade while in LTS.

mhdawson:
@shigeki still looking to get a first cut of this landed. Are you ok with landing or is there something that needs to be updated first.

mhdawson:
@shigeki ping ?

mhdawson:
@shigeki , @rvagg it would be good to get at least an initial version of this doc landed.  What needs to happen for that to be possible?

OpenSSL-Strategy.md
```diff
@@ -0,0 +1,94 @@
...
+
+Even if such support was to land in the Node.js codebase, the Node.js project will not commit to officially supporting LibreSSL or even maintaining full compatibility in an ongoing basis. As is the case for some operating systems that are not officially supported, ongoing compatibility will need to be monitored by interested parties who will also need to submit fixes as required.
+
+### [BoringSSL](https://boringssl.googlesource.com/boringssl/)
+
+> Although BoringSSL is an open source project, it is not intended for general use, as OpenSSL is. We don't recommend that third parties depend upon it. Doing so is likely to be frustrating because there are no guarantees of API or ABI stability.
+
+Google have made it clear that BoringSSL is not intended for general use outside of their own internal needs. Node.js will not officially support BoringSSL and unless trivially unobtrusive, the Node.js core team is discouraged from accepting changes that support BoringSSL. The appearance of support for BoringSSL will send the wrong message to users regarding its suitability as an OpenSSL replacement and Node.js' willingness to maintain support.
```
sam-github: I think @codebytere says Electron is using BoringSSL when they embed node. Why shouldn't our stance on any OpenSSL API-alike project be the same as for LibreSSL? No promises on support, but we'll accept PRs that don't burden Node.js in the maintenance of our own use cases.

tniessen: I actually talked to @codebytere at JSI and suggesred to bring a necessary change into Node.js core, so I am not generally opposed, but if we rephrase this, we should make it very clear that BoringSSL should not be depended upon unless absolutely necessary (which is the case for electron).

mhdawson:
@rvagg this has been opened for almost a year now.  Any chance we can get an update and get this landed?

OpenSSL-Strategy.md
sam-github: Re-reading, I think we are all saying the same thing, unless its a trivial unobtrusive change, we don't want to support BoringSSL. I'm OK with the original wording.

sam-github:
@rvagg do you want to get this landed? I'm happy to land it myself if you don't object (after taking another pass through to make sure the comments were addressed).

sam-github:
This is labelled `tsc-review`, did @nodejs/tsc review it?

rvagg:
@sam-github would you mind taking this over? It's had plenty of TSC review so far but needs some love to bring it up to date with current status (mostly your work).

I'd also be interested in handing over ownership of OpenSSL Evolution in the Strategic Initiatives https://github.com/nodejs/TSC/blob/master/Strategic-Initiatives.md#current-initiatives although I'm not sure it makes sense to put a non-TSC member in charge of it. Alternatively perhaps with this being landed it can be removed as a Strategic initiative.

mhdawson:
I don't think all of our champions need to be a TSC member, although maybe somebody who is the sponsor in the case where a non TSC member is doing most of the work.  The sponsor would be somebody that the champion can get help from if needed.

In terms of whether it needs to continue as a strategic initiative, once support for TLS 1.3 lands it may be done?

rvagg:
mm, the TLS 1.3 strategy isn't entirely clear wrt Node 11 and 10 so some finality on that would be good before taking this off the regular-reporting rotation. FIPS, {Boring,Libre}SSL, OpenSSL 2 all prevent this from going away entirely and it'd be good to have documentation and perhaps a person to query. I don't think I'm that person any more though.

sam-github:
@rvagg I'll update this a bit and push for landing it.

It may also be a good time/place to talk about whether to allow building against external OpenSSL that is NOT the same version as the on we use, I mentioned some issues coming up about that feature here: https://github.com/nodejs/node/pull/26319#issuecomment-473420410

rvagg:
@sam-github yeah that keeps on coming up and it's a legitimate concern, especially for Linux distribution packagers. IIRC I framed this doc to allow for it but only insofar as the core team are only going to concern themselves with the compatibility of the version bundled but will accept patches from outside to make things work. This is how it's always worked but there's been a blurry line when it involves a lot of changes (like that LibreSSL issue I jumped in to recently and probably shouldn't have!). So, as an example, if @kapouer / Debian wanted to build 12 against OpenSSL 1.1.0 for some reason and we didn't support it, they'd have to submit fixes to get it working again and the core team won't take responsibility for the compatibility issues that ensue (e.g. reports to nodejs/node about TLS 1.3 not working)—although in reality that'd be a case-by-case basis depending on who cared to respond to such a report.

If that's not clear in this doc already I think it should be made clear.

1. Core team cares mainly about directly bundled versions and will treat breakage & bugs with the same urgency as with any other core module
2. Node supports building against any other crypto library that's roughly compatible but creating that initial support and maintaining it will:
  a. be the responsibility of contributors (most likely not on the core team)
  b. not invoke the same kind of critical response to breakage
  c. be overridden or rejected where the change set might be deemed to be large or complex enough to get in the way of providing Node's core functionality and supporting it's primary dependency on OpenSSL

Or something like that. It would even allow for compiling master against 1.0.2 if someone cared enough to do that work and keep it updated and the code didn't step all over our toes.

sam-github:
@rvagg I've got time to integrate comments and update text, stuff like that, so I'm happy to help.

But... I thought I could push fixups to your branch, but I can't. Possibly your permissions, but more likely because I don't have push/merge perms on the TSC repo (I guess thats not surprising).

So, I'm not sure how I can help. I could just fork TSC, pull your branch from your repo, and PR it again, but I'm not sure duplicating the PR and forking the conversation will be helpful.

Any ideas?

rvagg:
I think this discussion has become stale in its current form so take my commits, edit and make this doc your own and open a new PR to start a fresh discussion. I think that'll be the best way.

sam-github:
I'll do that over the next day or so, thanks.

OpenSSL-Strategy.md
sam-github: resolved (only repo admins get a resolve button)

OpenSSL-Strategy.md
sam-github: resolved

OpenSSL-Strategy.md
```diff
@@ -0,0 +1,88 @@
...
+* [2eb170874](https://github.com/nodejs/node/commit/2eb170874aa5e84e71b62caab7ac9792fd59c10f): `openssl: fix keypress requirement in apps on win32`. A fix for the `openssl` client application, used by the Node.js test suite, to properly accept stdin without requiring a physical keyboard. As described in <http://openssl.6102.n7.nabble.com/PATCH-s-client-Fix-keypress-requirement-with-redirected-input-on-Windows-td46787.html>.
+* [664a65969](https://github.com/nodejs/node/commit/664a6596960655e214fef25e74d3285097703e95): `deps: add -no_rand_screen to openssl s_client`. Adds a `-no_rand_screen` command line option to the `openssl` client applicaiton, used by the Node.js test suite which skips invocation of the `RAND_screen()` call on Windows in order to speed up some of the Node.js TLS tests. Use of this can be found in many of the `test/*/test-{tls,https}-*` test files.
```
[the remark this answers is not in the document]
sam-github: resolved

OpenSSL-Strategy.md
sam-github: resolved

OpenSSL-Strategy.md
sam-github: 1.1.0 is EOL after 2019-09-11, resolved

OpenSSL-Strategy.md
sam-github: resolved

OpenSSL-Strategy.md
sam-github: resolved by updating BUILDING.md, see
- https://github.com/nodejs/node/pull/19794/commits/3f131489365dc83ba6adcdcc2a98fbdb6f63b4b5
- https://github.com/nodejs/node/pull/19794/commits/f584f354770ab44c79c1294f81a915f255d677e2
- https://github.com/nodejs/node/pull/25381/commits/5703c7f9a657b361e8ea5f8b6eb1174fe5598a64

OpenSSL-Strategy.md
sam-github: resolved like https://github.com/nodejs/TSC/pull/479#discussion_r268858758  I think, @shigeki, please comment if I misunderstand.

OpenSSL-Strategy.md
sam-github: resolved

OpenSSL-Strategy.md
sam-github: resolved

OpenSSL-Strategy.md
sam-github: resolved

OpenSSL-Strategy.md
sam-github: resolved

OpenSSL-Strategy.md
sam-github: Resolved: see https://github.com/nodejs/node/blob/master/deps/openssl/README.md which links to https://github.com/nodejs/node/blob/master/deps/openssl/config/README.md. I have sucessfully followed the documented process multiple times.

OpenSSL-Strategy.md
sam-github: resolved

OpenSSL-Strategy.md
sam-github: resolved

sam-github:
Replaced by https://github.com/nodejs/TSC/pull/677

PR #483, opened 2018-02-05
Add minutes for 2018-01-31 meeting with V8

hashseed:
@joyeecheung suggested that I should add this summary. So here it is.


mhdawson:
Landed as 5e237162855be622cfa5c66727282aa8db68554f

Issue #257, opened 2017-04-21
Approval to move node-heapdump into foundation


mhdawson:
Requesting approval to move github.com/bnoordhuis/node-heapdump to github.com/nodejs/node-heapdump

It been discussed here by the post-mortem group. Current proposal is to bring it in and then PR to accommodate changes/suggestions if they make sense (for example changing to use inspector protocol if that works)

There are 3 key artifacts for post-mortem diagnosis including:

* heapdumps
* first capture data (node-report)
* core dumps (llnode work is to allow working with those).

Bringing in node-heapdump would make sure we have heapdumps covered.  It seems widely used.

On approval we'd add in the Node.js governance and then transfer over.

jasnell:
+1 .... great to see this moving forward.

nebrius:
Forgive my ignorance, I haven't bumped into this project before, so my questions may have obvious answers. Is node-heapdump currently being used in the Node project itself as a dependency, or is it currently a standalone project? Along those lines, if we brought this into the foundation, would the long term goal be to include it in Node?

jasnell:
@nodejs/tsc ... please weigh in on this. I'd like to see if we can get consensus on this without calling for a formal vote.

@nodejs/post-mortem ... please see @nebrius' question above.

cjihrig:
I don't see a problem with bringing it into the foundation, but I'm curious why we need to.

misterdjules:
@nebrius 

>  Is node-heapdump currently being used in the Node project itself as a dependency, or is it currently a standalone project? 

It is a standalone project.

> Along those lines, if we brought this into the foundation, would the long term goal be to include it in Node?

I don't know. It's been pointed out by @bnoordhuis that V8 may soon expose an API that more or less provides the same facilities as node-heapdump. If/when that's the case, it's possible that it wouldn't necessarily make sense to bring node-heapdump into node's core, because the functionality would already be there. 

Not considering the possiblity of that V8 API, there have been discussions around integrating some of nodejs/node-report's functionality in core, including e.g generating heap dumps on out of memory errors, but so far I've been opposed to a lot of what has been suggested. I would think that if node-heapdump was integrated into core, then it could be used to generate these dumps. See https://github.com/nodejs/post-mortem/issues/44 for more context.

Disclaimer: I'm not too familiar with node-heapdump, or with the intentions of the people who advocate for integrating node-report in core, so please take my comments with a grain of salt. I'm just trying to provide more context.

[...]

PR #495, opened 2018-02-20
doc: update the repository management policy

joyeecheung:
Refs: https://github.com/nodejs/TSC/issues/257#issuecomment-366847601

cc @Trott

### DIFF
--- a/GitHub-Org-Management-Policy.md
+++ b/GitHub-Org-Management-Policy.md
@@ -59,8 +59,8 @@ Node.js Foundation GitHub Organization by opening an issue in the
 Provided there are no objections from any TSC or CommComm members raised in
 the issue, such requests are approved automatically after 72 hours. If any
 objection is made, the request may be moved to a vote in each of the
-Technical Steering and Community Committees. Both the TSC and CommComm must
-reject the request for it to be blocked.
+Technical Steering and Community Committees. If either the TSC or CommComm
+rejects the request, then the request is denied.
 
 In certain cases, Node.js Foundation Board of Directors approval may also be
 required.
--- a/GitHub-Org-Management-Policy.md
+++ b/GitHub-Org-Management-Policy.md
@@ -7,7 +7,7 @@ Community Committee (CommComm).
 
 ## Node.js Admin Repository
 
-The [Node.js admin repository](https://github.com/nodejs/admin) serves as the
+The [Node.js admin repository][nodejs/admin] serves as the
 central location for managing Node.js GitHub Organization administrative
 activities. Only Node.js GitHub Organization owners, TSC members, and Community
 Committee members have write permissions to the Node.js admin repository.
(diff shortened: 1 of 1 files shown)


GitHub-Org-Management-Policy.md
```diff
@@ -43,23 +43,27 @@ members to the organization when requested by a Working Group or team.
...
 
-No repository may be deleted, transferred into, or transferred out of the
-Node.js Foundation GitHub Organization without a simple majority of both the
-TSC and CommComm in favor of the action. In certain cases, Node.js Foundation
-Board of Directors approval may also be required.
+Any organization member may request the management of repositories within the
+Node.js Foundation GitHub Organization by opening an issue in the
+[Node.js admin repository][nodejs/admin]. The actions requested could be:
+
+- Creating a new repository
```
evanlucas: I had always thought that any member could create a repository... is that not the case?

joyeecheung: @evanlucas I believe so, just trying to document the process because we generally want people to open an issue before performing the actual action.

joyeecheung: Also, I am trying to document the whole process in https://github.com/nodejs/admin/pull/68

GitHub-Org-Management-Policy.md
```diff
@@ -43,23 +43,27 @@ members to the organization when requested by a Working Group or team.
...
+- Deleting an existing repository
+- Archiving an existing repository
+- Transferring a repository into or out of the organization
+
+Provided there are no objections from any TSC or CommComm members raised in
+the issue, such requests are approved automatically after 72 hours. If any
+objection is made, the request may be moved to a vote in each of the
+Technical Steering and Community Committees.  A simple majority of each group
```
Trott: While we're making this change, I'd opt to remove `simple majority of each group` and replace it with language that respects the decision making process of each group. For example, in the TSC, we do *not* use simple majority. We use absolute majority (discounting abstentions from the final total). Using `simple majority` here puts this document into conflict with TSC governance docs.

Maybe instead of "A simple majority of each group rejecting the request is required to block the request", it can be "Both the TSC and CommComm must reject the request for it to be blocked."

Also, should we consider blocking if *either* TSC or CommComm rejects? That's the way it is now and I think I kind of prefer it that way. Making things a little *easier* to block means you avoid hypothetical situations like this:

* Someone proposes creating `nodejs/foo`.
* TSC supports creating repo `nodejs/foo` and CommComm opposes it. It gets created.
* Someone else proposes deleting `nodejs/foo`
* TSC opposed deleting the repo and CommComm supports it. It gets deleted.
* Go to first bullet point. Repeat forever.

If only one or the other entity is enough to block, then things stay with whatever the status quo is and there's no endless add/delete/add/delete cycle. Admittedly, it's a hypothetical situation that may never occur, but perhaps a worthwhile thought experiment on the importance of deferring to the status quo when there is conflict?

gibfahn: >Also, should we consider blocking if either TSC or CommComm rejects? 

+1

joyeecheung:
Updated with @Trott 's suggestion. Good thought experiment :D

joyeecheung:
Will land this tomorrow if there were not objections.

GitHub-Org-Management-Policy.md
```diff
@@ -43,23 +43,27 @@ members to the organization when requested by a Working Group or team.
...
+- Creating a new repository
+- Deleting an existing repository
+- Archiving an existing repository
+- Transferring a repository into or out of the organization
+
+Provided there are no objections from any TSC or CommComm members raised in
+the issue, such requests are approved automatically after 72 hours. If any
+objection is made, the request may be moved to a vote in each of the
+Technical Steering and Community Committees. Both the TSC and CommComm must
+reject the request for it to be blocked.
```
gibfahn: I think we should consider @Trott 's second suggestion as well (https://github.com/nodejs/TSC/pull/495#discussion_r169428036):

> Also, should we consider blocking if *either* TSC or CommComm rejects? That's the way it is now and I think I kind of prefer it that way. Making things a little *easier* to block means you avoid hypothetical situations like this:
> 
> * Someone proposes creating `nodejs/foo`.
> * TSC supports creating repo `nodejs/foo` and CommComm opposes it. It gets created.
> * Someone else proposes deleting `nodejs/foo`
> * TSC opposed deleting the repo and CommComm supports it. It gets deleted.
> * Go to first bullet point. Repeat forever.
> 
> If only one or the other entity is enough to block, then things stay with whatever the status quo is and there's no endless add/delete/add/delete cycle. Admittedly, it's a hypothetical situation that may never occur, but perhaps a worthwhile thought experiment on the importance of deferring to the status quo when there is conflict?

Trott: Yes, I'd prefer that this sentence...:

> Both the TSC and CommComm must reject the request for it to be blocked.

...be changed to this:

> If either the TSC or CommComm rejects the request, then the request is denied.

richardlau:
The policy should really be moved over to https://github.com/nodejs/admin. We could then just have the document in this repo be a link.

joyeecheung:
@Trott @gibfahn Updated, PTAL.

@richardlau Yeah...I'll do that after this document is updated.

Trott:
Still LGTM!

PR #517, opened 2018-04-04
doc: add minutes for meeting 4 april 2018

mhdawson:


meetings/2018-04-04.md
```diff
@@ -0,0 +1,143 @@
...
+     Node.js was against them. Bradley/Trevor were the main ones speaking up but Bradley
+     has not been able to sync with Trevor to discuss. 
+  * Key issue was that changing storage also changes error handling.
+  * Trivial to implement zones with domains, but concerns were:
+     * error propagation issues
+       * you don’t know if your unhandled exception go
+       * with promises we already have post mortem issues from domains, people
+         Just making so errors get swallowed.
+    * concerns over performance
+      * Jeremiah, different perf concerns from async_hooks.  For async_hooks only
```
bmeurer: We've thought about zones for quite a bit when this was a hot topic, and I think the current proposal (although it's outdated) is fairly cheap even though it's always on. It's basically the cell to store the *current zone* plus an additional link for PromiseReactionJobs. To put this into perspective: `async_hooks` are **a lot** more expensive (on the promise side) when enabled.

PR #522, opened 2018-04-18
Update CTC to TSC

fhinkel:


fhinkel:
@ChALkeR There sure are other places 😄  Just changing this as I'm frequently copy&pasting it for the meetings.

ChALkeR:
@fhinkel Perhaps it would make sense to update https://github.com/nodejs/TSC/blob/master/README.md#dependent-projects in this PR too.

Issue #520, opened 2018-04-16
Node.js Foundation Technical Steering Committee (TSC) Meeting 2018-04-18


mhdawson:
## Time

**UTC Wed 18-Apr-2018 14:00 (02:00 PM)**:

| Timezone      | Date/Time             |
|---------------|-----------------------|
| US / Pacific  | Wed 18-Apr-2018 07:00 (07:00 AM) |
| US / Mountain | Wed 18-Apr-2018 08:00 (08:00 AM) |
| US / Central  | Wed 18-Apr-2018 09:00 (09:00 AM) |
| US / Eastern  | Wed 18-Apr-2018 10:00 (10:00 AM) |
| London        | Wed 18-Apr-2018 15:00 (03:00 PM) |
| Amsterdam     | Wed 18-Apr-2018 16:00 (04:00 PM) |
| Moscow        | Wed 18-Apr-2018 17:00 (05:00 PM) |
| Chennai       | Wed 18-Apr-2018 19:30 (07:30 PM) |
| Hangzhou      | Wed 18-Apr-2018 22:00 (10:00 PM) |
| Tokyo         | Wed 18-Apr-2018 23:00 (11:00 PM) |
| Sydney        | Thu 19-Apr-2018 00:00 (12:00 AM) |

Or in your local time:
* http://www.timeanddate.com/worldclock/fixedtime.html?msg=Node.js+Foundation+Technical%20Steering%20Committee%20(TSC)+Meeting+2018-04-18&iso=20180418T14
* or http://www.wolframalpha.com/input/?i=02PM+UTC%2C+Apr+18%2C+2018+in+local+time

## Links

* Minutes Google Doc: <https://docs.google.com/document/d/16i4zQXMdzxT-0j1vWe8t1967G37EKY09cQwEDn7yfww/edit>

## Agenda

Extracted from **tsc-agenda** labelled issues and pull requests from the **nodejs org** prior to the meeting.

### nodejs/node

* doc: add documentation on porting from Buffer(arg) usage [#19827](https://github.com/nodejs/node/issues/19827)
* Tracking Issue for Runtime Deprecation of Buffer constructor [#19079](https://github.com/nodejs/node/issues/19079)

### nodejs/TSC

* Tracking issue for updating TSC on Board Meetings [#476](https://github.com/nodejs/TSC/issues/476)
* Strategic Initiatives - Tracking Issue [#423](https://github.com/nodejs/TSC/issues/423)

### nodejs/admin

[...]

PR #523, opened 2018-04-18
Move Async Hooks to section that needs champions

fhinkel:
As discussed in TSC Meeting, https://github.com/nodejs/TSC/issues/520, move this to the section that needs champions.


Strategic-Initiatives.md
```diff
@@ -51,4 +50,4 @@ and have the support needed.
 [rvagg]: https://github.com/rvagg
 [trott]: https://github.com/trott
 [trevnorris]: https://github.com/trevnorris
```
cjihrig: I think this is the line that would be removed.

fhinkel: Thanks, didn't see that!

ofrobots:
I could help 'champion' async hooks. I can do so in a follow-on PR, or feel free to nominate me in this one :).

PR #545, opened 2018-05-21
remove GitHub policy

Trott:
The GitHub management policy now lives in the admin repository.


Trott:
Hmm, now that I'm coming back to this, probably best to replace it with a stub doc with a link to the new location, at least for a while.

targos:
I'm not sure it's worth it. The only reference to this file that I can found on Google is in https://github.com/nodejs/TSC/issues/490

Issue #492, opened 2018-02-12
Move make-node-meeting-artifacts into the foundation


gibfahn:
[mhdawson/make-node-meeting-artifacts](https://github.com/mhdawson/create-node-meeting-artifacts) is pretty useful, and is currently used in the TSC and Build repos (and probably others).

It may be too soon (I defer to @mhdawson on that) but we should probably think about bringing it into the foundation.

mcollina:
LGTM

mhdawson:
I had been meaning to open this issue had just not gotten to it yet :).  It is generating the artifacts for a number of WGs and the TSC and CommComm meetings and its time to move over.

After that happens the next step would be to look at moving over the automation I have to actually run at the right interval.

Trott:
Unfortunately, [the GitHub management policy](https://github.com/nodejs/TSC/blob/a942ad62a8c155010724ec2b12156ec284a64669/GitHub-Org-Management-Policy.md) still says this:

> No repository may be deleted, transferred into, or transferred out of the Node.js Foundation GitHub Organization without a simple majority of both the TSC and CommComm in favor of the action. 

It also strongly implies that this needs to go in the admin repo rather than the TSC repo.

I would strongly support a PR that altered policy to reflect our current practice. If we want to change that practice, let's actually road-test the change before documenting (and forgetting) it (because forgetting it is what you do if you don't actually *do* it for a while first--people much more strongly take their cues from what everyone else is doing than from documented policies).

jasnell:
SGTM. The most likely best home for this, however, is the @nodejs/automation WG

Fishrock123:
Sounds good to me, and also agree about putting it under the automation team.

Trott:
Just to be clear: 

**This cannot happen without a TSC vote and a CommComm vote.**

(See https://github.com/nodejs/TSC/issues/492#issuecomment-366113635 if you're not sure why that is. IMO, that portion of the policy is misguided and should be changed, but for now, it's the policy.)

fhinkel:
Ping @nodejs/tsc and @nodejs/community-committee. Please vote whether you're in favor of moving the meeting tooling to the foundation, see https://github.com/nodejs/TSC/issues/492#issuecomment-366113635. 
Please use thumbs up/down on the first post.

mhdawson:
SGTM just to make my vote clear.

chowdhurian:
SGTM, on putting the tooling under the stewardship of the automation WG

[...]

PR #591, opened 2018-08-27
doc: change links in ONBOARDING.md

gdams:
I was reading through the ONBOARDING guide and noticed that it was pointing to @mhdawson's fork. Please close if this was intentional.


richardlau:
For some history the link is pointing to where the tool was originally until it was brought under the foundation (https://github.com/nodejs/TSC/issues/492) so I think this was just an oversight.

PR #595, opened 2018-09-05
doc: add minutes for meeting Sep 9

mhdawson:


mhdawson:
#nodejs/TSC any reviewers?

ChALkeR:
The commit description looks incorrect — there wasn't a meeting on 2018-09-09.

mhdawson:
@ChALkeR seems like I got that wrong, but not sure how to fix now.

mhdawson:
The name of the file and the info in the file does match the correct date Sep 5

ChALkeR:
@mhdawson I don't think any action is needed, as all the files themself seem correct.
But I left a comment at the commit just in case.

PR #677, opened 2019-03-25
doc: describe OpenSSL maintenance strategy

sam-github:
This replaces https://github.com/nodejs/TSC/pull/479

### DIFF
--- a/OpenSSL-Strategy.md
+++ b/OpenSSL-Strategy.md
@@ -1,0 +1,356 @@
+# Node.js OpenSSL Strategy
+
+This policy document describes for each release line:
+
+* Which versions of OpenSSL Node.js will include.
+* Which versions of a "shared" OpenSSL Node.js will support.
+* What version of TLS will be supported
+  - by default,
+  - with run-time configuration.
+* Whether FIPS will be supported.
+
+It also gives background on OpenSSL release lifetimes, TLS1.3, and FIPS support,
+as they affect Node.js.
+
+
+## Node.js version-specific strategy
+
+### Node.js version 4.x (EOL 2018-04-30)
+
+No longer maintained.  Not discussed further.
+
+
+### Node.js versions 6.x (EOL 2019-04-30) and 8.x (EOL 2019-12-31)
+
+Node.js 6.x and 8.x have the same OpenSSL versions and policies.
+
+Node.js 8.x is unusual because it's end-of-life would normally be April, 2020,
+but it was moved earlier to conincide with the end-of-life of OpenSSL 1.0.2.
+
+* OpenSSL version: 1.0.2 (kept up-to-date with latest)
+* Allowed shared OpenSSL version: 1.0.2x
+* Default minimum TLS version is TLSv1, default maximum is TLSv1.2. No other
+  TLS versions are supported by this version of OpenSSL or Node.js.
+* FIPS: supports being configured to build against OpenSSL FIPS 2.0. FIPS
+  configuration is not default, and OpenSSL FIPS 2.0 is not included in
+  Node.js.
+
+
+### Node.js version 10.x (EOL April-2021)
+
+OpenSSL LTS support timing, the lack of OpenSSL LTS planning and the lack of a
+clear timeframe for a new FIPS module complicates Node.js 10.
+
+For Node.js < 10.16.0:
+* OpenSSL version: 1.1.0
+* Allowed shared OpenSSL version: 1.1.0
+* Default minimum TLS version is TLSv1, default maximum is TLSv1.2. No other
+  TLS versions are supported by this version of OpenSSL or Node.js.
+* FIPS: not supported
+
+For Node.js >= 10.16.0:
+* OpenSSL version: 1.1.1
+* Allowed shared OpenSSL version: 1.1.0 or 1.1.1
+* Default minimum TLS version is TLSv1, default maximum is TLSv1.2. It is
+  expected that TLS1.3 support will be backported to 10.x, but it will *not* be
+  supported by default, only by explicit run-time configuration.
+* FIPS: not supported
+
+The plan described above is to:
+
+* Remain backward-compatible with OpenSSL 1.1.0 via dynamic linking for the
+  lifetime of Node.js 10. This will cease being an _official_ policy at the
+  OpenSSL 1.1.0 EOL date (2019-09-11), which will occur long before Node.js 10.x
+  enters Maintenance LTS (April 2020) or EOL (April 2021). Node.js does _not
+  officially support_ EOL versions of OpenSSL, but support for a shared OpenSSL
+  1.1.0 will be maintained as long as it does not cause security problems and
+  there are contributors available to maintain such support. The lack of a
+  passing test suite with Node.js 10 compiled against OpenSSL 1.1.0 past the
+  OpenSSL EOL date will not hold up further releases of Node 10.
+* Upgrade to OpenSSL 1.1.1 when made generally available and Node.js 10 can
+  retain ABI and API compatibility. OpenSSL 1.1.1 is claimed by the OpenSSL
+  project to be API and ABI compatible with 1.1.0, and Node.js experiments
+  indicates that it is, see
+  https://github.com/nodejs/node/issues/18770#issuecomment-446785733.
+* **No support for FIPS** The latest [OpenSSL 3.0 and FIPS Update][] does not
+  predict that an OpenSSL 1.1.1 ABI compatible library with FIPS support will
+  ever exist. Without such a library, Node.js 10.x cannot support FIPS.
+
+This plan must be communicated to users of Node.js 10 early and often.  There is
+potential for instability and a change in default OpenSSL version is
+unprecedented and therefore unexpected. The potential for breaking API and/or
+ABI may also cause disruption, potentially requiring an increment of
+`NODE_MODULE_VERSION`, which will also be unprecedented within a single release
+line. There is no indication yet that this will happen when OpenSSL 1.1.1 is
+included in Node.js 10.x, but it is important that users be aware of this
+possibility.
+
+
+### Node.js version 11.x (EOL June-2019)
+
+This release will not be designated LTS. It was updated to include OpenSSL 1.1.1
+in [11.9.0](https://nodejs.org/en/blog/release/v11.9.0/).
+
+For Node.js < 11.9.0:
+* OpenSSL version: 1.1.0
+* Allowed shared OpenSSL version: 1.1.0
+* Default minimum TLS version is TLSv1, default maximum is TLSv1.2. No other
+  TLS versions are supported by this version of OpenSSL or Node.js.
+* FIPS: not supported
+
+For Node.js >= 11.9.0:
+* OpenSSL version: 1.1.0
+* Allowed shared OpenSSL version: 1.1.0 or 1.1.1
+* Default minimum TLS version is TLSv1, default maximum is TLSv1.2. It is
+  expected that TLS1.3 support will be backported to 11.x, but it will *not* be
+  supported by default, only by explicit run-time configuration.
+* FIPS: not supported
+
+
+### Node.js version master, 12.x (release expected April-2019, EOL April-2022)
+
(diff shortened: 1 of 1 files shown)


sam-github:
I wish I made this a draft PR, but I don't seem able to change it after the fact.

I updated some text with current information about OpenSSL reease strategies, since OpenSSL 1.1.1 has been released, and has been designated LTS and support until 2023 has been committed to.

I have started to add information about OpenSSL 3.0.0 and FIPS.

I have not yet added information about TLS1.3 strategy.

All of which is to say, probably not worth reviewing yet.

sam-github:
@shigeki re: https://github.com/nodejs/TSC/pull/479#issuecomment-369622355

You bring up a couple issues:

1. I believe you already updated the assembler  requirements in node.js master, and this doesn't require mentioning in the policy document. Agreed?

2. I am not familiar with the OpenSSL extensions described. The plan ATM is to land https://github.com/nodejs/node/pull/26270, so 10.x's assembler requirements will become identical to 11.x (and master, and 12.x). Do you, or @nodejs/tsc @nodejs/crypto , see a problem with this?


> If we add a new feature specific to 1.1.1, the API should check openssl version and be noted in the API doc

Maybe we _should_ do this, but I don't believe we _have_ to. Shared OpenSSL libraries are not supported by us, they do not necessarily support Node.js' APIs, they are probably not floating our patches, they may be several patch releases behind the version of OpenSSL we include, so have bugs our Node.js builds do not, and be missing features we have. I think we should continue to make a best effort to support shared openssll libraries, including feature detection, but I don't think we are required to document the effect of shared library builds in our documentation. For our own sanity, we probably need to keep the unit tests running against shared library builds, either against 1.1.0, or 1.1.1a. TLS1.3 won't work with OpenSSL < 1.1.1b, for example, and the Ed curves won't work with OpenSSL 1.1.0.

I'm in the middle of backporting TLS1.3 support to 11.x, and in the process I expect to run into some of the issues with shared library builds.

sam-github:
@shigeki and sorry, I forgot

> For the API simplicity, I think it is better to support one release version of OpenSSL during LTS.

I'm not sure exactly what you mean, is it that:
1. you object to updating OpenSSL from 1.1.0 to 1.1.1 during the 10.x lifetime (I don't think that's what you mean)
2. once we update to 1.1.1 in 10.x, you suggest we stop supporting 1.1.0 as a shared library (this is very attractive, if possble)
3. for 12.x and future LTS, we do not support shared OpenSSL libs that aren't the same version as "our" OpenSSL (this has happened already on master, 12.x won't build against 1.1.0).

OpenSSL-Strategy.md
```diff
@@ -0,0 +1,361 @@
...
+  override this using a `with-openssl` option which will compile against the
+  version of OpenSSL that was most recently installed with `brew` but this is
+  not believed to be in common use.
+
+
+## Node.js version-specific strategy
+
+### Node.js version 4.x (EOL 2018-04-30)
+
+No longer maintained.  Not discussed further.
```
sam-github: I removed the discussion of 4.x, it doesn't seem relevant.

OpenSSL-Strategy.md
```diff
@@ -0,0 +1,361 @@
...
+
+* OpenSSL version: 1.0.2 (kept up-to-date with latest)
+* Allowed shared OpenSSL version: 1.0.2x
+* Default minimum TLS version is TLSv1, default maximum is TLSv1.2. No other
+  TLS versions are supported by this version of OpenSSL or Node.js.
+* FIPS: supports being configured to build against OpenSSL FIPS 2.0. FIPS
+  configuration is not default, and OpenSSL FIPS 2.0 is not included in
+  Node.js.
+
+XXX Do we support FIPS builds, or do we treat like other shared builds, best
```
sam-github: I think we support FIPS more than "normal" shared library builds, but I'm not certain.

OpenSSL-Strategy.md
```diff
@@ -0,0 +1,361 @@
...
+
+For Node.js >= 10.16.0:
+* OpenSSL version: 1.1.1
+* Allowed shared OpenSSL version: 1.0.2 or 1.1.0 or 1.1.1
+* Default minimum TLS version is TLSv1, default maximum is TLSv1.2. It is
+  expected that TLS1.3 support will be backported to 10.x, but it will *not* be
+  supported by default, only by explicit run-time configuration.
+* FIPS: not supported
+
+XXX Check whether our test matrix is updated to test this.
```
sam-github: I'll check this.

OpenSSL-Strategy.md
```diff
@@ -0,0 +1,361 @@
...
+XXX Node.js 10.x allows building against an external OpenSSL 1.0.2, so doesn't
```
sam-github: I haven't tried (yet) but the `./configure` option of 10.x sure makes it look like it supports building against FIPS.

OpenSSL-Strategy.md
```diff
@@ -0,0 +1,361 @@
...
+For Node.js < 10.16.0:
+* OpenSSL version: 1.1.0
+* Allowed shared OpenSSL version: 1.0.2 or 1.1.0
+* Default minimum TLS version is TLSv1, default maximum is TLSv1.2. No other
+  TLS versions are supported by this version of OpenSSL or Node.js.
+* FIPS: not supported
+
+For Node.js >= 10.16.0:
+* OpenSSL version: 1.1.1
+* Allowed shared OpenSSL version: 1.0.2 or 1.1.0 or 1.1.1
```
sam-github: I have to confirm that 1.0.2 is possible. I assume that any shared builds that worked with 10.15.3 have to work with 10.16.0?

OpenSSL-Strategy.md
```diff
@@ -0,0 +1,361 @@
...
+
+For Node.js >= 11.9.0:
+* OpenSSL version: 1.1.0
+* Allowed shared OpenSSL version: 1.0.2 or 1.1.0 or 1.1.1
+* Default minimum TLS version is TLSv1, default maximum is TLSv1.2. It is
+  expected that TLS1.3 support will be backported to 11.x, but it will *not* be
+  supported by default, only by explicit run-time configuration.
+* FIPS: not supported
+
+XXX Confirm what versions of shared OpenSSL are expected to be supported.
```
sam-github: I will confirm what shared openssl versions we supported before 11.9.0 (when openssl 1.1.1 was integrated), and after 11.9.0.

OpenSSL-Strategy.md
```diff
@@ -0,0 +1,361 @@
...
+
+### OpenSSL 1.1.1
+
+The OpenSSL team has designated 1.1.1 the next LTS line and have made a
+commitment that it will remain both API and ABI compatible with OpenSSL 1.1.0,
+and will be supported until 2023-09-11.  It is the first OpenSSL version to
+support TLS 1.3, however Node.js' TLS1.3 support requires at least OpenSSL
+1.1.1b.
+
+### OpenSSL 3.0.0
```
sam-github: This entire section is new, and has some bad news now that the OpenSSL project has released more information about their FIPS support plan, PTAL.

OpenSSL-Strategy.md
mhdawson: We treated failures in the FIPs builds the same as any other failure, should be blocking.

OpenSSL-Strategy.md
mhdawson: I don't think we should break 1.0.2 support in 10.x.  If its a matter or not being able to get 1.1.1 in without doing that then we need to discuss.

mhdawson:
This is a great doc, let us know when it is ready to be reviewed/land.

It is unfortunately related to FIPS, but there is not much we can do.

OpenSSL-Strategy.md
sam-github: Confirmed 1.0.2 isn't supported by v10.x. 1.1.0 is. I updated text.

OpenSSL-Strategy.md
sam-github: confirmed, 1.0.2 is not supported, 1.1.0 is

OpenSSL-Strategy.md
sam-github: If 10.x supports 1.0.2 then we should continue to support it. I built v10.x in ci, https://ci.nodejs.org/job/node-test-commit-linux-containered/11905/, and it built and passed against 1.0.2r, but my local attempts to build against shared openssl 1.0.2q fail completely, way too many missing symbols. I'm retrying with openssl 1.0.2r, though since patch releases shouldn't add new APIs, I'm mystified.

richardlau: > 
> 
> If 10.x supports 1.0.2 then we should continue to support it. I built v10.x in ci, https://ci.nodejs.org/job/node-test-commit-linux-containered/11905/, and it built and passed against 1.0.2r, but my local attempts to build against shared openssl 1.0.2q fail completely, way too many missing symbols. I'm retrying with openssl 1.0.2r, though since patch releases shouldn't add new APIs, I'm mystified.

er the build you've linked to skips over openssl 1.0.2 (and 1.1.1 and fips):
image

sam-github: No, I can't build against 1.0.2r either. I'm retrying with a different openssl build configuration. @rvagg any comment? I even grepped around, AFAICT node_crypo.cc uses symbols not defined by openssl 1.0.2. I might need access to the docker containers to see what's going on.

OpenSSL-Strategy.md
```diff
@@ -107,6 +107,28 @@ For Node.js >= 10.16.0:
   supported by default, only by explicit run-time configuration.
 * FIPS: not supported
 
+XXX
+
+https://ci.nodejs.org/job/node-test-commit-linux-containered/11905/
+https://ci.nodejs.org/job/node-test-commit-linux-containered/nodes=ubuntu1604_sharedlibs_openssl102_x64/11899/consoleText
+
+makes it seem 1.0.2 is supported, but I can't reproduce that build.
```
richardlau: @sam-github https://ci.nodejs.org/job/node-test-commit-linux-containered/nodes=ubuntu1604_sharedlibs_openssl102_x64/11899/ is from a Node.js 6 build:

image


https://ci.nodejs.org/job/node-test-commit-linux-containered/11905/ skipped building against openssl 1.0.2, openssl 1.1.1 and fips.

sam-github: @richardlau So, when I look at https://ci.nodejs.org/job/node-test-commit-linux-containered/11905/, and its a v10.x build, all the light green checkmarks at the bottom are platforms that were skipped? Wow. And thanks!

On the plus side, that confirms that I'm not totally insane, and that openssl 1.0.2 is not in fact supported by node 10.x.

OpenSSL-Strategy.md
sam-github: It does not, because it won't build against 1.0.2

OpenSSL-Strategy.md
richardlau: > 
> 
> @richardlau So, when I look at https://ci.nodejs.org/job/node-test-commit-linux-containered/11905/, and its a v10.x build, all the light green checkmarks at the bottom are platforms that were skipped? Wow. And thanks!
> 

Correct. The faded light green is skipped.

sam-github:
OK, @nodejs/crypto @nodejs/tsc I believe this describes current policy. Not that it can't change later :-).

Note in particular that 12.x will never support FIPS, because FIPS will require openssl 3.0.0 which will NOT be ABI compatible with openssl 1.1.1.

13.x Might support FIPS, depends on when its cut (its not in the release timeline yet)

So, there is pretty much guaranteed to be "FIPS gap". Not too sure what we can do about that. I suggest once Openssl 3.0.0 FIPS is closer to reality that we revisit, to see if there are any options.

sam-github:
@nodejs/tsc PTAL

https://github.com/sam-github/TSC/blob/openssl-strategy/OpenSSL-Strategy.md#nodejs-version-specific-strategy is the actual policy, the following sections are background information on Node.js and OpenSSL, for those interested.

ChALkeR:
Belated LGTM.

PR #678, opened 2019-03-27
Replace @jasnell as Champion for core promises API

mcollina:
I'm happy to be the Champion for the core promises API.

cc @jasnell


Trott:
The link for jasnell's profile (in the list at the bottom) is now unused. Can you remove it?

mcollina:
> The link for jasnell's profile (in the list at the bottom) is now unused. Can you remove it?

@Trott done.

PR #681, opened 2019-03-31
Remove joyee from champions for error migration

joyeecheung:
I have not been very active on this initiative for a while and will be focusing on other stuff in the foreseeable future so I think I should just remove myself as a champion. If anyone would like to become a new champion for this feel free to push a new commit here or open a follow on PR.

Although IMO the initiative is no longer strategic for us anymore as it's mostly done and the remaining unmigrated errors are all fairly rare. Maybe a more reasonable move is to create a separate section for these inactive (or no longer strategic?) initiatives, or move it to completed (which feels a bit odd since technically it is not completed).


gireeshpunathil:
+1 to not designating it a `strategic initiative` as the major part of the initiative is completed.

PR #685, opened 2019-04-02
add strategic initiative for Build resources

Trott:
From https://github.com/nodejs/build/issues/1154#issuecomment-448418977:

---

Rich and I (Rod) will begin a strategic initiative sometime in January
to do some accounting of resources and costs (people and other) and try
and come up with proposals and plans for reducing pain and improving the
experience for collaborators while managing quality. Or something like
that. Basically we need to think more strategically about this and
understand what our problems and risks are and what we can do about
them.

---

This is that initiative (in April rather than January, but such is
life).

### DIFF
--- a/Strategic-Initiatives.md
+++ b/Strategic-Initiatives.md
@@ -11,20 +11,21 @@ and have the support needed.
 
 # Current Initiatives
 
-| Initiative        | Champion                                                  | Links                                                            |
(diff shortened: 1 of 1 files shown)


Trott:
(Use "hide whitespace" in the diffs to make this manageable. I added whitespace on all lines to have stuff line up in a monospace font.)

PR #705, opened 2019-05-15
updates to strategic initiatives

Trott:
* Move OpenSSL Evolution to complete
* Move Async Hooks to needs-champion
* Remove Rod (but leave Rich) for Build Resources


mhdawson:
@sam-github I think creating a new one around FIPS would make sense.  If there is an issue you can point to volunteer to be the champion.

PR #763, opened 2019-10-03
remove/update references to Node.js Foundation in README

Trott:
In some cases, the right thing to do was change it to say OpenJS
Foundation. In other cases, the right thing to do was to change it to
say Node.js.

@nodejs/tsc PTAL


Trott:
There are definitely further updates to come IMO, but this seems like a good (and hopefully non-controversial) start.

PR #799, opened 2020-01-08
doc: describe security workflow

sam-github:
/cc @mralekzandr @nodejs/security

### DIFF
--- a/Security-Team.md
+++ b/Security-Team.md
@@ -106,3 +106,59 @@ the Node.js program on HackerOne.
 - [@watson](https://github.com/watson) - Thomas Watson
 
 <!-- ncu-team-sync end -->
+
+## Security Triage Workflow
+
+Security issues are reported via HackerOne, see [nodejs.org](https://nodejs.org/en/security/#reporting-a-bug-in-node-js).
+
+Report states are described in HackerOne [docs](https://docs.hackerone.com/programs/report-states.html).
+
+The expected workflow for issues reported to Node.js is:
+
+1. `NEW`: issue reported (initial state).  Issues in this state are in the
+   process of being analyzed with the intention to move them into either
+   `TRIAGED`, `NEEDS-MORE-INFO`, or one of the "Closed" states
+   (`NOT-APPLICABLE`, `INFORMATIVE`, etc.). Has several sub-states:
+   1. `H1 Triage`: Issues start off assigned to the `H1 Triage` team. That team
+      does initial triage. They ask for a runnable reproduction, they run it,
+      and they verify whether it does in fact appear to demonstrate a
+      vulnerability that has not already been reported. Once this is done, they
+      assign the issue to the `Node.js Team`.
+   2. `Node.js Team`: Issues assigned by `H1 Triage`, or issues that Node.js
+      assigns to itself for triage preemptively (perhaps because of priority, or
+      because its disposition is obvious, or any other reason we decide to
+      skip the `H1 Team`'s reproduction).
+   3. `NEEDS-MORE-INFO`: In the state of back-and-forth with reporter.
+      When further progress on the issue is blocked on response from the reporter,
+      this is the state.
+2. `TRIAGED`: State for an issue when it is a non-duplicate and resolvable.
+   Issues in this state are waiting for the Node.js team to take some action to
+   resolve them. Typically, this would be publishing a Node.js release that
+   resolves the issue, but it could also be publishing documentation, or making
+   a configuration change to our infrastructure for a https://nodejs.org
+   problem. The expected resolution should be described in a comment on the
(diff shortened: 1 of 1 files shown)


mralekzandr:
@sam-github - I forgot to mention this in my email! Disclosing Informative or N/A reports can help deter other researchers from submitting similar reports, thus potentially cutting down on future informative or n/a reports! It gives folks an idea of things you are NOT looking for.

sam-github:
@mralekzandr That's my hope as well, if we build up a list of things we don't consider issues (like anything using `process.binding()`), it can help people avoid wasting time researching or reporting them.

Security-Team.md
```diff
@@ -106,3 +106,60 @@ the Node.js program on HackerOne.
...
+     getting resolved. Comments on the issue when someone is working on it are
+     useful, as is assignment of the issue to a person who agress to be
+     responsible for making sure it is resolved.
+5. `RESOLVED`: State for an issue that has had a fix published. Issues in this
+   state should be disclosed.
+6. ...: Final states for issues that we will not fix:
+   - `NOT-APPLICABLE`: We do not agree this a vulnerability. We can request
+     that the reporter close it to preserve their H1 "reputation points".  If
+     its an issue, but not a vulnerability, we can suggest that it be reported
+     to the Node.js issue tracker. ***Should it be disclosed?***
```
mhdawson: Are these ***Should it be disclosed*** questions?

sam-github: yes

mhdawson: Story still a bit confused.  So this is a WIP versus a doc ready to land?

sam-github: these lines don't exist anymore

mhdawson:
@sam-github thanks.

sam-github:
@mralekzandr added some new features, see below. I'll incorporate into these docs.

I've created 3 new queues for your inbox. To lessen the manual work required by your team, I created inbox views (instead of groups), so reports will automatically funnel to each queue based on where in the life cycle the report sits. (Note - I couldn't use "Pending Disclosure Decision" because of a 20 character limit in naming new inbox views/queues.)

 Can you either a) relay this information to the rest of the team, or b) point me toward the right github issue to give the update?

Review 4 Disclosure - ALL closed reports (except for Spam) in which neither the reporter nor your team have requested disclosure.

Disclosure Requested - ALL closed reports (except for Spam) in which either the reporter or your team have requested disclosure. Pending confirmation by the other party.

Does it make sense to separate this into 2 queues - once in which the reporter has requested disclosure and we are waiting for your team to confirm, and another in which your team has requested disclosure and we are waiting on the reporter to confirm?

Disclosed Reports - All disclosed reports

sam-github:
@mralekzandr How do I move things from the `Review 4 Disclosure` queue to a `Decided not to disclose` queue after they have been reviewed?

mralekzandr:
@sam-github Thanks for your patience while I mulled this over.

Given some current UI limitations, we're going to have to get creative with a "Decided Not To Disclose" queue.

I noticed that you guys don't use the "Reference" field for reports (well, there is one report with the reference field filled out, but it appears to be a possible mistake). We could add a `Without References` filter to the "Review 4 Disclosure" queue. Then, when you've decided a report isn't eligible for disclosure, you can add "DON'T DISCLOSE" to that reports `Reference` field, which would then pull it out of the "Review 4 Disclosure" queue.

Simultaneously, we'd create a new queue called "Will Not Disclose", using the same filters as "Review 4 Disclosure" queue, but ADD `With References`as a filter. 

So the workflow from the Review 4 Disclosure queue would be:

1. If you want to disclose, request disclosure and follow that workflow (the report would then move from "Review 4 Disclosure" --> "Pending Disclsoure"
2. If you do NOT want to disclose, you'd add "DON"T DISCLOSE" to the reference field (the report would then move from "Review 4 Disclosure" --> "Will Not Disclose"


My sincerest apologies if this isn't entirely clear. I know it's a wonky work around, but I think it's the best option we have given your current workflows and our own platform limitations.

I'm happy to hop on a call next week to walk you through it - sometimes a screen share is easier to relay information. I haven't implemented anything I mentioned here, so let me know your thoughts and we'll go from there.

Thanks!
-Alek

sam-github:
@mralekzandr Lets not go through that much trouble. I documented that reports should be considered for disclosure when they are closed. Lets leave it at that.

The pre-existing state where we hadn't kept up to date with disclosure made it look like a way of managing that would be useful, but going forward, our report rate is low enough that I think we can just ensure we consider disclosure as reports are closed.

@nodejs/tsc PTAL I think this is ready to merge.

Marked up version: https://github.com/nodejs/TSC/blob/h1-process/Security-Team.md#security-triage-workflow

PR #812, opened 2020-02-05
Remove Workers from strategic initiatives

addaleax:
Workers have been stable for a while now, and while there are still PRs being made against the Worker code, the activity does not stand out from other Node.js core areas and I don’t think the feature requires any special handling at this point.

Also, my name was misspelled.


addaleax:
@trott Sure, I’ve added it there.

PR #833, opened 2020-03-06
doc: update strategic initiatives

mhdawson:
* Move N-API to complete
* Move Standards to complete
* Move New Stream APIs to needs mentor
* Remove Mentoring from needs volunters as owned by CommComm
* Remove nsp and modules as modules has an entry and not sure what other
  this was covering


mhdawson:
@joyeecheung we talked about moving the Standards one to complete due to the coverage in the OpenJS level. Let me know if we should keep that and I'll add it back in.

@gabrielschulhof I've moved N-API to complete as a strategic initiative, let me know if you think we should keep it instead.

mhdawson:
@gabrielschulhof would you like to chime in before I land as well.

PR #859, opened 2020-05-01
Update OpenSSL Strategy for OpenSSL 3.x

sam-github:


sam-github:
@danbev @jasnell The state as I know it. The deadlines as I understand them!

OpenSSL-Strategy.md
```diff
@@ -117,12 +117,55 @@ For Node.js >= 11.9.0:
...
+3.x will be backported to it.
+
+## Node.js version 16.x (est Apr 2021) (EOL Apr 2024)
+
+* OpenSSL version: 3.x (because of OpenSSL EOL)
+* Allowed shared OpenSSL version: 1.1.1 (proposed...)
+* Default minimum TLS version is TLSv1.2, default maximum is TLSv1.3. TLSv1
+  and TLSv1.1 are *not* supported by default, only by explicit run-time
+  configuration.
+* FIPS: not supported
```
sam-github, line 146: or, maybe supported? it could happen.

jasnell: Current expectation, as I understand it, is that the FIPS provider *should* be available by the end of the year

mhdawson: Even if it comes for a later version in the 3.x as long as it's not breaking we should be able to get to FIPs as well. Therefore, it is more of a "maybe" at this point.

OpenSSL-Strategy.md
```diff
@@ -117,12 +117,55 @@ For Node.js >= 11.9.0:
...
+
+OpenSSL 1.1.1 goes EOL on 2023-09-11, which is before 16.x will go EOL, so is
+not an appropriate choice for 16.x
+
+For minimal disruption, it would probably be helpful if Node.js supported
+building against OpenSSL 1.1.1 out-of-tree, even if OpenSSL 3.x was in-tree.
```
mhdawson, line 152: Agreed

Issue #864, opened 2020-05-11
Collaboration and Contribution Process Review


bnb:
I briefly reached out to @mcollina about this who suggested I open an issue on the TSC repo:

I'm not sure if we've ever done a full review/reflective pass on our collaboration and contribution documents within the project (for example, the [TSC Charter](https://github.com/nodejs/TSC/blob/master/TSC-Charter.md) and [Working Groups](https://github.com/nodejs/TSC/blob/master/WORKING_GROUPS.md) doc,).

It may perhaps be beneficial to collectively take a look at these processes (and any others!) and see both which parts may or may not be reflective of how we currently operate and which parts could be improved to allow us both as a project to work better *and* for the community to contribute effectively and meaningfully.

Trott:
I think we're overdue to move whatever is of value in `Applications` elsewhere and then remove the directory in this repository. We don't have projects applying for membership anymore. They should go to the OpenJS Foundation instead, I would think.

jasnell:
It's going to be easiest to start with a list of issues that may need to be addressed, otherwise it's going to be way too open ended.

mcollina:
@bnb would you like to faciltiate this? I think having an approach from _outside_ the folks with Node.js commit access would be very interesting.

boneskull:
(also collaborators who _would_ contribute more often, if not for the process. 🙂)

bnb:
@mcollina sure. Is this something you'd want me to do inside of the TSC repo, inside of the Admin repo, or somewhere else?

mcollina:
I think opening issues here is ok.

Having a fresh perspective on https://github.com/nodejs/node/blob/master/CONTRIBUTING.md and how we could simplify things.

Trott:
> Having a fresh perspective on https://github.com/nodejs/node/blob/master/CONTRIBUTING.md and how we could simplify things.

Since most of us who are frequent contributors to the project also never look at that file, it's possible there's some pretty low-hanging fruit in there.  My immediate suggestions:

* Simplify language everywhere. Be brief, informative, and friendly.
* For issues:
  * Remove "How to Contribute in Issues".
  * Remove "Resolving a Bug Report". 
* For pull requests:
  * Remove the preamble that is before the table of contents.
[...]

PR #865, opened 2020-05-14
Proposal: Node.js Contributor Survey

jasnell:
Building on the discussion that has been started in https://github.com/nodejs/TSC/issues/864, I'd like to propose that we take a queue from other significant open source projects and conduct our own Contributor Survey -- with the intent of helping us to determine where we can continue to make improvements to our contribution process.

Using the 2019 Kubernetes Contributor Survey as an initial starting point, I've put together an initial set of questions that can allow us to start structuring something.

This is a draft. The intent here is to build / refine the list of questions we'd like to see on a survey should we decide to do one.

Please do not yet merge this PR. It is still a work in progress

/cc @bnb

### DIFF
--- a/surveys/contributor/2020/2020-Contributor-Survey.md
+++ b/surveys/contributor/2020/2020-Contributor-Survey.md
@@ -1,0 +1,120 @@
+# DRAFT -- 2020 Node.js Contributor Survey
+
+* How long have you been contributing to Node.js?
+
+  "Contribution" here means much more than just code. Participation in
+  issue and PR discussions, contributions to the Community Committee,
+  the website, the documentation, and so forth.
+
+  * Less than one year
+  * One to two years
+  * Two to three years
+  * Three to five years
+  * Five+ years
+
+* What kind of contributor do you consider yourself to be?
+
+  "Contribution" here means much more than just code. Participation in
+  issue and PR discussions, contributions to the Community Committee,
+  the website, the documentation, and so forth.
+
+  * I'm not yet a contributor
+  * I've contributed once or twice
+  * I occasionally contribute
+  * I regularly contribute
+  * I've contributed substantially
+  * I used to contribute but no longer do
+
+* Over the past year, do you consider your contributions to have:
+
+  * Increased
+  * Decreased
+  * Remained about the same
+
+* In the coming year, do you expect your contributions to:
+
+  * Increase
+  * Decrease
+  * Remain the same
+  * I have no idea
+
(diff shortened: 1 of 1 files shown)


2020-Contributor-Survey.md
```diff
@@ -0,0 +1,124 @@
+# DRAFT -- 2020 Node.js Contributor Survey
+
+* How long have you been contributing to Node.js?
```
ljharb, line 3: Can we open by defining “contribution” to make sure it doesn’t just include code?

2020-Contributor-Survey.md
```diff
@@ -0,0 +1,124 @@
...
+  * Remain the same
+  * I have no idea
+
+* What region of the world are you in?
+
+* How many other open source projects not under the Node.js project do you contribute to?
+
+  * This is my first open source project!
+  * 1 other
+  * 2 or more
```
ljharb, line 39: this is a broad category :-) maybe add one more tranche?

2020-Contributor-Survey.md
jasnell: Absolutely.

2020-Contributor-Survey.md
jasnell: Heh... I was thinking about dropping this question. Not sure it adds much value. I think a more compelling question would be whether the individual contributes to other projects in the ecosystem.

mcollina:
Good work!

bnb:
I would also like to say that we should probably have a survey for folks outside of our current collaborator base to understand what would make Node.js a better environment for them to collaborate and ensure we’re not reinforcing our own biases ❤️

mmarchini:
Small suggestion: move the survey to `surveys/contributor/` folder.

jasnell:
The idea of doing a contributor survey was raised at either the last TSC meeting or the one before (I can't recall exactly). There were no objections so I am moving forward with this. I am currently putting the survey draft together on survey monkey and will post links to the survey once it is ready to go. As there have been no further suggestions on questions, I'll be going with what has been discussed already unless there are objections

mhdawson:
@jasnell my guess is that this now can be closed as the survey is complete? Discussion of the results seems to be in the PR with those results.

mhdawson:
Or I guess landed since it's a PR. @jasnell does this match what was sent out, if so we should just convert from "work in progress" and land.

jasnell:
This can be closed

PR #877, opened 2020-06-05
Fix broken link to release team

richardlau:
The existing link is dead (since https://github.com/nodejs/node/pull/23927 renamed the section).

It looks like it is more accurate to say @nodejs/releasers (i.e. the team that can do releases) rather than link to the signing keys.

### DIFF
--- a/Security-Team.md
+++ b/Security-Team.md
@@ -16,7 +16,7 @@ The policy for inclusion is as follows:
 
 1. All members of @nodejs/TSC have access to private security reports and
    private patches.
-2. Members of the [release team](https://github.com/nodejs/node#release-keys)
+2. Members of the @nodejs/releasers team
    have access to private security patches in order to produce releases.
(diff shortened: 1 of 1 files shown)


Security-Team.md
```diff
@@ -16,7 +16,7 @@ The policy for inclusion is as follows:
 
 1. All members of @nodejs/TSC have access to private security reports and
    private patches.
-2. Members of the [release team](https://github.com/nodejs/node#release-team)
+2. Members of the [release team](https://github.com/nodejs/node#release-keys)
```
gireeshpunathil, line 19: ya - given https://github.com/orgs/nodejs/teams/releasers exists, I guess it looks like a better choice to me!

richardlau:
```suggestion
2. Members of the @nodejs/releasers team
```

richardlau: Changed.

PR #882, opened 2020-06-14
2020 Node.js Contributors Survey Results

jasnell:
@nodejs/tsc @nodejs/collaborators ... the compiled results of the 2020 Contributor's Survey.


ronag:
That was some interesting reading. Seems to me like there are a lot of mentions of the contribution & review process.

I'm rather new to the organization but how are these results usually consumed? i.e. do they lead directly to any actionable items or are they used more as reference for ongoing and future discussions?

addaleax:
So … yup, this is definitely very interesting. What’s a good place to discuss the results? Here, in this PR?

One thing that comes up quite a bit is being annoyed with C++, which is more than understandable. That was obviously my main motivation behind adding `src/README.md`, but it seems like that’s either not enough or not what people are looking for, or it’s not discoverable enough. I’m wondering if any @nodejs/collaborators could expand on that, or would be willing to chat about it with me (or others)? I know I do a lot of C++ work, so I do feel that I have a responsible to make that work as accessible to others as possible.

ronag:
> One thing that comes up quite a bit is being annoyed with C++, which is more than understandable. 

I've got around 14 years of experience with C++ from working on https://github.com/CasparCG/server (Windows). However, I have not touched C++ in NodeJS core at all. Not sure if/how I can be of help but if you need someone to experiment with in how to make the C++ parts more approachable I'd be happy to contribute.

Qard:
> One thing that comes up quite a bit is being annoyed with C++, which is more than understandable. That was obviously my main motivation behind adding `src/README.md`, but it seems like that’s either not enough or not what people are looking for, or it’s not discoverable enough. I’m wondering if any @nodejs/collaborators could expand on that, or would be willing to chat about it with me (or others)? I know I do a lot of C++ work, so I do feel that I have a responsible to make that work as accessible to others as possible.

I think `src/README.md` is an excellent resource but, at least for me, I've found it challenging figuring out _where_ exactly things are in the source files, what the class hierarchy is, and awareness of helpers or "better" ways to do things such as using `MaybeStackBuffer` rather than `std::vector`. I learn about a lot of these things from suggestions in code review, which is fine, but it _does_ mean more iterations which can overwhelm new contributors. I go in _expecting_ a certain level of iteration because I've been in the issue tracker long enough to have set my expectations but and new contributor might not be expecting so much iteration and interpret it negatively.

Flarna:
Regarding C++: `src/README.md` provides a good overview/starting point for people able to speak C++. It's really good as the v8 API itself is not that easy to use/undrestand and the docs there are not that helpful.

I'm old enough that I grow up with C and later C++. During the last years I worked with quite some younger people - mostly fresh from university. They did great work JavaScript/Java/... and associated toolings but once I confronted them with tasks in C++ area there were lost.

So I fear the issue is more the language C++ itself not how it is used by node.

What I miss is some more description about the interworking between JS/C++ which are not that obvious. For example the various places where AliasedBuffer/... are used.  It's not that easy to track down all cases where C++ influences JS as you can't just follow a call tree.

mmarchini:
Ultra-minor nit on formatting:

> * "There are a few individuals that are extremely noisy and write complex comments to grasp. Their language is sophisticated and they are hard to follow for non-native english speakers.
> * They are making Node.js seem an exclusive club of people who have a lot of time to contribute."

Seems like those two items should be only one:

> * "There are a few individuals that are extremely noisy and write complex comments to grasp. Their language is sophisticated and they are hard to follow for non-native english speakers. They are making Node.js seem an exclusive club of people who have a lot of time to contribute."

jasnell:
@mmarchini ... I split those intentionally so the second point wouldn't get lost as it's an important point on it's own.

jasnell:
Going to merge this but I don't want to discourage further discussion and review

bnb:
In the future, perhaps this could live in nodejs/admin ❤️

PR #919, opened 2020-09-05
feat(charter): introduce emergency and async votes

mmarchini:
Asynchronous voting is appropriate for issues that should be votes by
as many TSC members as possible.

Emergency votes are intended as a tool to resolve issues that are
time-sensitive and TSC can't reach consensus on it in a timely manner
asynchronously.

### DIFF
--- a/TSC-Charter.md
+++ b/TSC-Charter.md
@@ -139,9 +139,9 @@ object?" as a final call for dissent from the consensus.
 
 If an agenda item cannot reach a consensus a TSC member can call for
 either a closing vote or a vote to table the issue to the next meeting or to
-be conducted asynchronously through a system to be chosen and seconded by the
-majority of TSC members. The call for a vote must be seconded by a majority
-of the TSC or else the discussion will continue.
+be conducted asynchronously through a system to be chosen by the TSC. The
+call for a vote is approved if none of the TSC members in the meeting object
+to it.
 
 For all votes, a simple majority of all TSC members for, or against, the issue
 wins. A TSC member may choose to participate in any vote through abstention.
@@ -149,16 +149,16 @@ wins. A TSC member may choose to participate in any vote through abstention.
 Note that, in addition to requiring a simple majority vote of the TSC, all
 changes to this charter are also subject to approval from the CPC.
 
-### Section 7.1. Emergency Votes
+### Section 7.1. Immediate Votes
 
 For time-sensitive issues where reaching consensus asynchronously fails, TSC
-members may call for a vote through the TSC mailing list. A compelling reason
+members may call for a vote through the TSC issue tracker. A compelling reason
 must be provided on why calling for a vote can't wait until the next meeting.
-Emergency Votes must be seconded by a majority of the TSC, otherwise the
-issue will be discussed in the next meeting (where a regolar vote can be
-called).
+The issue must remain open for 24 hours before starting the vote process. If
+there are any objections to having a vote, the issue will be discussed in the
+next meeting (where a regular vote can be called).
 
-Except for how they are called, emergency votes follow the same rules as
+Except for how they are called, immediate votes follow the same rules as
 regular votes.
 
 ## Section 8. Project Roles
--- a/TSC-Charter.md
+++ b/TSC-Charter.md
@@ -138,10 +138,10 @@ item has appeared to reach a consensus the moderator will ask "Does anyone
 object?" as a final call for dissent from the consensus.
 
 If an agenda item cannot reach a consensus a TSC member can call for
-either a closing vote or a vote to table the issue to the next meeting or to
-be conducted asynchronously through a system to be chosen by the TSC. The
-call for a vote is approved if none of the TSC members in the meeting object
-to it.
+a closing vote (assumes enough TSC members in the meeting), a vote to table the
+issue to the next meeting, or a vote to be conducted asynchronously through a
+system to be chosen by the TSC. The call for a vote is approved if none of the
+TSC members in the meeting object to it.
 
 For all votes, a simple majority of all TSC members for, or against, the issue
 wins. A TSC member may choose to participate in any vote through abstention.
--- a/TSC-Charter.md
+++ b/TSC-Charter.md
@@ -68,7 +68,7 @@ participate in TSC discussions, *and* does not participate in TSC votes, the
 member shall be automatically removed from the TSC. The member may be invited
 to continue attending TSC meetings as an observer.
 
-## Section 4. Responsibilities of the TSC.  
+## Section 4. Responsibilities of the TSC.
 
 Subject to such policies as may be set by the CPC, the TSC is
 responsible for all technical development within the Node.js project,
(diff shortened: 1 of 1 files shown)


mcollina:
I would prefer to call them "immediate votes".

mmarchini:
I went with "emergency" because I thought it was a more familiar terminology, but I'm happy with "immediate" as well

TSC-Charter.md
```diff
@@ -138,16 +138,29 @@ item has appeared to reach a consensus the moderator will ask "Does anyone
...
 Note that, in addition to requiring a simple majority vote of the TSC, all
 changes to this charter are also subject to approval from the CPC.
 
+### Section 7.1. Emergency Votes
+
+For time-sensitive issues where reaching consensus asynchronously fails, TSC
+members may call for a vote through the TSC mailing list. A compelling reason
+must be provided on why calling for a vote can't wait until the next meeting.
+Emergency Votes must be seconded by a majority of the TSC, otherwise the
+issue will be discussed in the next meeting (where a regolar vote can be
```
targos, line 158:
```suggestion
issue will be discussed in the next meeting (where a regular vote can be
```

MylesBorins:
Quick note that charter changes will need to be approved by the CPC. One thought. Could we maybe make the language in the charter around votes more ambiguous and move the specifics of implementation to the governance? That way if we want to iterate on this new process we can do so without having to go back-and-forth with the CPC

TSC-Charter.md
```diff
@@ -138,16 +138,29 @@ item has appeared to reach a consensus the moderator will ask "Does anyone
...
+majority of TSC members. The call for a vote must be seconded by a majority
+of the TSC or else the discussion will continue.
 
 For all votes, a simple majority of all TSC members for, or against, the issue
 wins. A TSC member may choose to participate in any vote through abstention.
 
 Note that, in addition to requiring a simple majority vote of the TSC, all
 changes to this charter are also subject to approval from the CPC.
 
+### Section 7.1. Emergency Votes
```
mhdawson, line 152: My personal feeling is it might be better to just allow votes to be called through the mailing list.  I'm not sure a meeting is needed, and the mailing is likely to reach more TSC members.

MylesBorins: +1 to making a clear process for calling a vote without requiring a meeting rather than making a specific carve out. Honestly if one TSC member calls the vote and there are no objections within a 24 hour period that could be reasonable enough to kick one off. Waiting for a majority of the TSC to +1 could end up causing a delay long enough that we have the next meetings ready to go already 😂

mmarchini: So instead of having two ways of calling for vote and executing a vote, we only have one via mailing list (potentially via issues for votes that can be public)? I think that's a good idea, it allows for input from any TSC members (not only those who live in a timezone favourable for a particular meeting) and allow us to vote faster.

mmarchini: @MylesBorins +100000, there's no reason to wait for 10 +1s via email to start the vote process, 24h with no one objecting to having a vote seems reasonable.

MylesBorins: If someone doesn't like the vote they can vote no 😂

mcollina: I agree! I would like to see the need for calling a vote in a meeting or with a majority removed.

Trott:
Just to make sure everyone is aware: Changes to the charter need to be approved by the foundation board. So let's figure out what we want soon because there's a second set of people that need to concur!

Trott:
> Quick note that charter changes will need to be approved by the CPC.

Oh, I thought it was the board. CPC makes more sense, I think.

MylesBorins:
@Trott it use to be the board until the OpenJS foundation was formed. After OpenJS started the CPC was chartered with the responsibility of overseeing project charters and the Node.js TSC + CommComm were rechartered under the CPC.

IMHO this is a massive benefit to the new structure as we have far more representation on the CPC and they meet far more regularly.

mhdawson:
@mmarchini do you want to update this PR to reflect what seems to be the consensus?

mmarchini:
yes, i just haven't had time to do it yet. If someone else wants to do it though, feel free to do so

mmarchini:
Ok, I made the following changes:

- Call for votes follows Lazy Consensus now
- Renamed "Emergency Votes" to "Immediate Votes"
- Calling an immediate vote must be done in the issue tracker, and not via mailing list (so the call is public)

While I like the idea of merging both votes and moving more vote details out of the charter to our governance, I think it can be done in another PR. This PR should considerably improve the situation already.

mmarchini:
cc @nodejs/tsc as we need majority approval for charter changes. Once we approved it at TSC level I'll raise it to CPC.

TSC-Charter.md
```diff
@@ -138,16 +138,29 @@ item has appeared to reach a consensus the moderator will ask "Does anyone
...
 For all votes, a simple majority of all TSC members for, or against, the issue
 wins. A TSC member may choose to participate in any vote through abstention.
 
 Note that, in addition to requiring a simple majority vote of the TSC, all
 changes to this charter are also subject to approval from the CPC.
 
+### Section 7.1. Immediate Votes
+
+For time-sensitive issues where reaching consensus asynchronously fails, TSC
+members may call for a vote through the TSC issue tracker. A compelling reason
```
mcollina, line 155: There have been a few occasions where the issue to vote on was private in nature. In those cases, we should use emails.

TSC-Charter.md
```diff
@@ -138,16 +138,29 @@ item has appeared to reach a consensus the moderator will ask "Does anyone
 object?" as a final call for dissent from the consensus.
 
 If an agenda item cannot reach a consensus a TSC member can call for
-either a closing vote or a vote to table the issue to the next meeting.
-The call for a vote must be seconded by a majority of the TSC or else the
-discussion will continue.
+either a closing vote or a vote to table the issue to the next meeting or to
+be conducted asynchronously through a system to be chosen by the TSC. The
+call for a vote is approved if none of the TSC members in the meeting object
+to it.
```
mhdawson, line 144:
```suggestion
a closing vote (assumes enough TSC members in the meeting), a vote to table the
issue to the next meeting, or a vote to be conducted asynchronously through a
system to be chosen by the TSC. The call for a vote is approved if none of the
TSC members in the meeting object to it.
```

mhdawson (approved):
LGTM

TSC-Charter.md
```diff
@@ -138,16 +138,29 @@ item has appeared to reach a consensus the moderator will ask "Does anyone
...
+system to be chosen by the TSC. The call for a vote is approved if none of the
+TSC members in the meeting object to it.
 
 For all votes, a simple majority of all TSC members for, or against, the issue
 wins. A TSC member may choose to participate in any vote through abstention.
 
 Note that, in addition to requiring a simple majority vote of the TSC, all
 changes to this charter are also subject to approval from the CPC.
 
+### Section 7.1. Immediate Votes
```
apapirovski, line 152: Wouldn't time-sensitive be more apt (already used in describing the process below)? The phrase `immediate` seems pretty confusing to me, especially since the process is still not truly immediate.

mcollina: Yes

joesepi (approved):

bnb:
+1. Don't want to approve directly since it'll green check and this isn't a space where I should be green checking ❤️

WaleedAshraf (approved):

bnb:
Any progress on this?

bnb:
removed the cc-agenda label for now. Feel free to add it back for us when this is active/ready to merge again <3

Trott:
Given that this has been open for more than a year and has seen no activity for more than 7 months, I'm going to close this. But if you feel that's the wrong thing to do and there's still a need for pursuing this change, please re-open. Thanks!

PR #945, opened 2020-10-26
doc: avoid directing to HTTP if possible

tniessen:
There is no reason to use plain HTTP, it only makes things easier to track/hijack for ISPs and other untrustworthy entities.


tniessen:
There is one other HTTP link in `OpenSSL-Strategy.md`, but it doesn't work over HTTPS.

Trott (approved):

targos (approved):

PR #1097, opened 2021-10-11
Adjust charter to simplify automated removal rules

Trott:
This is a charter change, so it would require CPC approval.

The reason for this change is that the current automatic removal rules are unworkable. We currently have (I believe) 5 TSC members who do not meet the 25% meeting attendance requirement. Of those 5, at least one has participated in a vote, leaving us with 4 TSC members who may be eligible for supposedly-automatic removal. However, "does not participate in TSC discussions" is impossible to automate or even be certain about. I suspect that with that criteria, only 1 of the TSC members is eligible for automatic removal. But even that, I can't be sure. And it's awkward to bring it up. That's why it's supposed to be automatic! Because we're so bad at this!

So, I propose two changes to make this a possible-to-automate process similar to the automated Collaborator-removal process we have now in the main repo.

1. Eliminate the impossible-to-automate "participate in TSC discussions" loophole. That's what this change does.
2. Start keeping track of votes in documents that are kept in this repository, similar to the way we have meeting attendees documented in the meeting minutes. There are possible variations on this, but the point is that there is one source of truth about who did and did not participate in votes.

Once those two changes are done, we can look at automating the process so that it is actually automatic in reality, and not merely on paper.

@nodejs/tsc

### DIFF
--- a/TSC-Charter.md
+++ b/TSC-Charter.md
@@ -62,7 +62,8 @@ through accessible public postings.
 
 TSC members are expected to regularly participate in TSC activities.
 
-A TSC member is automatically removed from the TSC if, during a 3-month period, all of the following are true:
+A TSC member is automatically removed from the TSC if, during a 3-month period,
+all of the following are true:
 
 * They attend fewer than 25% of the regularly scheduled meetings.
 * They do not participate in any TSC votes.
@@ -82,7 +83,7 @@ including:
 * Maintaining the list of additional Collaborators.
 * Development process and any coding standards.
 * Mediating technical conflicts between Collaborators or Foundation
-projects.
+  projects.
 
 The TSC will define Node.js project’s release vehicles.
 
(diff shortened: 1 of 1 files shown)


apapirovski:
Strictly speaking you could take engagement in the TSC repo, or on TSC PRs, as a sign of the "participate in TSC discussions discussions" criteria — or just reword it to mean that.

apapirovski:
> If a meeting didn't happen, it didn't happen. We still seem to be running around 8 or 9 meetings per 3 month period, which seems like plenty to me.

Yeah, I'm all good with the reasoning personally. Mostly just pointing out potential loopholes that could still arise. I've always thought this whole process should be more automatable. (And I say that as someone that has been extremely bad at actually attending meetings. Sigh.)

targos:
The change LGTM.

> Start keeping track of votes in documents that are kept in this repository

+1. Even better if it's in a format that can be easily read by a computer.

Trott:
> LGTM. Maybe we should call out that abstaining from a vote still counts as participating.

I'm happy to do that. I didn't do it initially to keep things concise and because abstentions are specifified as a way to _participate_ previously in the charter:

> A TSC member may choose to participate in any vote through abstention.

PR #1114, opened 2021-11-08
update list of users with HackerOne access

Trott:


Security-Team.md
```diff
@@ -51,16 +51,16 @@ includes a fix, documentation, an informational CVE or blog post.
...
+* [kumarak](https://github.com/kumarak) - **Akshay K**
+* [mike-myers-tob](https://github.com/mike-myers-tob) - **Mike Myers**
+* [tn3rt](https://github.com/tn3rt) - **Trent Brunson**
 * [vdeturckheim](https://github.com/vdeturckheim) - **Vladimir de Turckheim**
-* [@watson](https://github.com/watson) - **Thomas Watson**
+* [woodruffw-tob](https://github.com/woodruffw) - **William Woodruff**
+* [xlauko](https://github.com/xlauko) - **Heinrich Lauko**
 
-List is from the [member page](https://hackerone.com/nodejs/team\_members) for
+The list is from the [member page](https://hackerone.com/nodejs/team\_members) for
```
targos, line 63: Who has access to this page? It doesn't exist for me.

mcollina (approved):
lgtm

Security-Team.md
Trott: @targos Remove the backslash escape from the page and it will hopefully exist for you. If not, let me know and I'll look at what groups you are in in HackerOne and compare them to what's needed to see that page.

Trott: (And by the way, the backslash escape is removed in https://github.com/nodejs/TSC/pull/1115.

targos: I already tried without the backslash.

targos: image

mhdawson (approved):
LGTM

Security-Team.md
Trott: @targos I guess that means you need to be an Admin to see the page. I've made you a HackerOne Admin. Let me know if you can see the page now.

I have some additional thoughts on this that I'll take to the TSC email list.

PR #1118, opened 2021-11-12
add initial vote data

Trott:
This is a final step before being able to automate the identification of
inactive TSC members per the charter.

### DIFF
--- a/votes/2021-09-29-0.json
+++ b/votes/2021-09-29-0.json
@@ -1,0 +1,18 @@
+{
+  "description": "Moving an inactive collaborator to emeritus",
+  "refs": ["https://github.com/nodejs/node/pull/40115#issuecomment-930656531"],
+  "votes": {
+    "Trott": 1,
+    "ronag": 1,
+    "mcollina": 1,
+    "apapirovski": 1,
+    "aduh95": 1,
+    "cjihrig": 1,
+    "targos": 1,
+    "BridgeAR": 1,
+    "fhinkel": 1,
+    "jasnell": 1,
+    "mhdawson": 1,
+    "gireeshpunathil": 0
+  }
+}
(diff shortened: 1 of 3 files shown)


MylesBorins (approved):
LGTM

richardlau:
> 
> 
> This is a final step before being able to automate the identification of inactive TSC members per the charter.

Might there also need to be some way of identifying who the TSC members were at the time of the votes? For example, I wasn't on the TSC for these votes.

cjihrig (approved):

richardlau (approved):

Trott:
> > This is a final step before being able to automate the identification of inactive TSC members per the charter.
> 
> Might there also need to be some way of identifying who the TSC members were at the time of the votes? For example, I wasn't on the TSC for these votes.

The script I have ready to go takes care of that. It checks who the TSC members are today and then checks out README.md only from 3 months ago and compares. Anyone not on both lists is ignored. So it wouldn't bother trying to figure out if you had voted or not.

votes/2021-09-29-0.json
```diff
@@ -0,0 +1,18 @@
...
+    "mcollina": 1,
+    "apapirovski": 1,
+    "aduh95": 1,
+    "cjihrig": 1,
+    "targos": 1,
+    "BridgeAR": 1,
+    "fhinkel": 1,
+    "jasnell": 1,
+    "mhdawson": 1,
+    "gireeshpunathil": 0
```
gireeshpunathil, line 16: checking at the referenced issue comment, I see that I have abstained from voting for the said topic. just want to understand, is that what `0` means here?

gireeshpunathil: ah, forget the comment, I see it is described below. thanks!

gireeshpunathil (approved):

jasnell (approved):

PR #1119, opened 2021-11-13
chore: fix minor grammar typo in Security-Team.md

Trott:
There is a missing _is_ in the existing sentence, but this rewords it a bit.

### DIFF
--- a/Security-Team.md
+++ b/Security-Team.md
@@ -134,7 +134,7 @@ The expected workflow for issues reported to Node.js is:
 3. `RESOLVED`: State for an issue that has had a fix published. Issues in this
    state should be disclosed.
 4. ...: Final states for issues that we will not fix:
-   * `NOT-APPLICABLE`: We do not agree this a vulnerability. We can request
+   * `NOT-APPLICABLE`: The issue reported is not a vulnerability. We can request
      that the reporter close it to preserve their H1 "reputation points".  If
      it is an issue, though not a vulnerability, we can suggest that it be
      reported to the Node.js issue tracker.


aduh95 (approved):

jasnell (approved):

Security-Team.md
```diff
@@ -134,7 +134,7 @@ The expected workflow for issues reported to Node.js is:
 3. `RESOLVED`: State for an issue that has had a fix published. Issues in this
    state should be disclosed.
 4. ...: Final states for issues that we will not fix:
-   * `NOT-APPLICABLE`: We do not agree this a vulnerability. We can request
+   * `NOT-APPLICABLE`: The issue reported is not a vulnerability. We can request
```
DerekNonGeneric, line 137:
```suggestion
   * `NOT-APPLICABLE`: The issue reported is not a vulnerability. We can request
```

The previous wording makes more practical sense to me. Who is to say what _is_ or _ins't_ a &ldquo;vulnerability&rdquo;? Is that decided using the consensus-seeking model implemented privately amongst the members of this team? If there is an aspect of this team's work that is time-sensitive, perhaps &ldquo;[Lazy Consensus](https://github.com/nodejs/TSC/blob/119a2cf0e1308b2841063d3dac6f5900e6f432ea/TSC-Charter.md?plain=1#L131-L134)&rdquo; may not be the most suitable operating mode, which seems to have been the status-quo.

I am also concerned that if any of the other ~100 core collaborators feel so inclined to fix a random &ldquo;security&rdquo; (or gray-area) problem that happens to turn up, that we should have to follow some additional protocol or process (e.g., _responsible disclosure_) beyond continuing to work in public and addressing any such problems/concerns via our PR + review process (as we have been doing this whole time).

Trott: @DerekNonGeneric This text is inapplicable to almost all collaborators. This is how private security vulnerability reports are triaged in HackerOne. Only a small number of collaborators have access to these reports.

PR #1139, opened 2021-12-15
chore: add vote data for automatic collaborator move-to-emeritus

Trott:
Refs: https://github.com/nodejs/node/pull/41155

### DIFF
--- a/votes/2021-12-15-0.json
+++ b/votes/2021-12-15-0.json
@@ -1,0 +1,18 @@
+{
+  "description": "Automatic move-to-emeritus for inactive collaborators",
+  "refs": ["https://github.com/nodejs/node/pull/41155"],
+  "votes": {
+    "Trott": 1,
+    "mcollina": 1,
+    "jasnell": 1,
+    "MylesBorins": 1,
+    "targos": 1,
+    "tniessen": 1,
+    "joyeecheung": 1,
+    "richardlau": 1,
+    "danielleadams": 1,
+    "BridgeAR": 1,
+    "mhdawson": 1,
+    "aduh95": 1
+  }
+}


richardlau (approved):

mhdawson (approved):
LGTM

cjihrig (changes requested):
It was not clear to me that this was a vote. Apparently [I wasn't the only one](https://github.com/nodejs/node/pull/41155#issuecomment-995039849).

Trott:
> It was not clear to me that this was a vote. Apparently [I wasn't the only one](https://github.com/nodejs/node/pull/41155#issuecomment-995039849).

It wasn't officially, but I figure that erring on the side of considering "waiting for a majority of TSC to approve" type things TSC votes is preferable. The TSC Charter puts the baseline for TSC participation as participating in one vote in a three month period, so the more things that are considered votes, the easier it is to meet that requirement. Non-participation in a vote is not penalized, so counting more things as votes doesn't hurt anyone. That's not the case for meetings, where there is a percentage attendance threshold.

If you still think this shouldn't land, I won't argue and will close it.

cjihrig:
> Non-participation in a vote is not penalized, so counting more things as votes doesn't hurt anyone.

I think it's a very unlikely scenario, but couldn't this technically become the only vote in a 3 month period?

I won't block this, but I really think we should be more explicit moving forward about what constitutes a vote since it can technically impact someone.

Trott:
> I think it's a very unlikely scenario, but couldn't this technically become the only vote in a 3 month period?

Sure but in that case, leaving this out so that there are *no* votes in a 3-month period would mean that only attendance at meetings matters. Adding a vote in would make it easier to meet the either/or requirement of vote or attendance. With no votes, you have to be attending the meetings or else you don't meet the requirements.

Trott:
> I won't block this, 

I'm going to close this. It wasn't a vote strictly speaking and I don't want to do something that can be questioned/debated if there's a -0 in the mix. 

> but I really think we should be more explicit moving forward about what constitutes a vote

I can't argue with that.

targos:
I assumed a change to the governance document required a vote but I'm probably wrong

Trott:
> I assumed a change to the governance document required a vote but I'm probably wrong

The usual consensus process is sufficient. We can't change the TSC charter that way,  but governance of Node.js repo is ordinary decision-making.

We could choose to either informally agree or formally require that such changes go through a TSC vote.

Issue #1207, opened 2022-04-07
request: update OpenSSL-Strategy.md w/ strategy on Node 17/18


johndevedu:
Please update https://github.com/nodejs/TSC/blob/main/OpenSSL-Strategy.md with strategy on Node.js version 17.x and 18.x. Since the statically linked OpenSSL has been updated from 1.1.1 to 3.0.0+quic (https://github.com/nodejs/node/issues/40106 https://github.com/nodejs/node/pull/38512), it will be helpful to know what the plans are for supporting dynamically linked 1.1.1 in 17.x and 18.x moving forward. 

Also, there were challenges mentioned in [Node 16 + OpenSSL strategy](https://github.com/nodejs/TSC/blob/main/OpenSSL-Strategy.md#nodejs-version-16x-est-apr-2021-eol-apr-2024) (i.e. algorithms in legacy lib, etc), so it would be helpful to know how these challenges will be/have been overcome. 

TIA!

mhdawson:
@johndevedu the doc's been updated, is that what you were looking for in terms of the strategy for 18.x?

mhdawson:
I'm going to close this since it was updated, and we've not had a response from the op for a few months. Please let us know if that was not the right thing to do.

johndevedu:
I'm sorry for the late reply. Yes, that was exactly what was needed. Thank you!!!

PR #1211, opened 2022-04-14
doc: update outdated OpenSSL 3.x strategy

danbev:
Refs: https://github.com/nodejs/TSC/issues/1207

### DIFF
--- a/OpenSSL-Strategy.md
+++ b/OpenSSL-Strategy.md
@@ -151,14 +151,9 @@ of 15.x (which is only about 8 months), that experimental support for OpenSSL
 * Default minimum TLS version is TLSv1.2, default maximum is TLSv1.3. TLSv1
   and TLSv1.1 are _not_ supported by default, only by explicit run-time
   configuration.
-* FIPS: unpredictable, see below
 
 OpenSSL 1.1.1 goes EOL on 2023-09-11, which is before 16.x will go EOL, so is
 not an appropriate choice for 16.x
-
-OpenSSL 3.x may have FIPS support by release date of Node.js 16.x, or it may
-not. If it does not, since its a provider, it may be possible to support in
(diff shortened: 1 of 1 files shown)


Trott (approved):

tniessen (approved):

BethGriggs (approved):

richardlau:
> Refs: #1207

The ask in https://github.com/nodejs/TSC/issues/1207 was to update the strategy for Node.js 17/18, and in particular the state of continued support for dynamically linking OpenSSL 1.1.1. My own view there is that we could say we support that for as long as OpenSSL 1.1.1 is in support.

Note that OpenSSL 1.1.1 is scheduled to go out of support after 2023-09-11 [1] which is before the End-of-Life dates for Node.js 18, but also Node.js 16 and we have to make a decision at some point regarding what we're going to do with Node.js 16. We have precedent before of shortening the expected lifespan of a Node.js LTS release (Node.js 8) to coincide with the end of support for the OpenSSL version it was built against. If we chose to do that we need to make that decision earlier and communicate that to the wider ecosystem.

(The strategy doc says
> OpenSSL 1.1.1 goes EOL on 2023-09-11, which is before 16.x will go EOL, so is not an appropriate choice for 16.x

but OpenSSL 3.0.0 wasn't available when Node.js 16.x went out so we had no choice but to release on OpenSSL 1.1.1.)

We are using the quic fork of OpenSSL 1.1.1 -- I haven't seen any indication that they would differ regarding how long 1.1.1 would be supported for.


[1] https://www.openssl.org/policies/releasestrat.html

danbev:
> The ask in #1207 was to update the strategy for Node.js 17/1

This PR intentionally does not address the strategy, that is the reason for using `Refs` instread of  `Fixes`. Sorry if that was confusing.

cjihrig (approved):

mhdawson (approved):
LGTM

Issue #1201, opened 2022-04-01
2022 TSC Chair election - Planning for May 1 ballot.


mhdawson:
May is the month for TSC Elections. The proposed schedule is:

April 16 - Call for nominations
April 30 - Close nominations
May 1 - Open balloting
May 15 - Announce Results
May 16-29- Handover
May 30 (or earlier if handover complete) Transition complete

Sound good?

Trott:
This is fine by me, but I alos wouldn't mind a shorter timeframe. Like, I don't think we need to wait two weeks before opening the call for nominations, and I don't think we need to leave the call open for two weeks, and I don't think the balloting needs to be open for two weeks. All of those things strike me as "one week is plenty of time" kind of things.

jasnell:
+1 to @Trott's suggestion for one week for each step.

Trott:
Like I said, this is fine with me, but I would prefer this:

* April 4: Call for Nominations
* April 10: Close nominations
* April 11: Open balloting
* April 17: Close balloting
* April 18 Announce results and begin handover (if any is needed)

gireeshpunathil:
looking at [last election](https://github.com/nodejs/TSC/issues/1013) for prior arts:
 - the current chair's tenure ends on May 30th, so any new schedule schould aim for transition by that date, not before?
 - there was a discussion on multiple chairs / vice chair that was not pursued then, so if we stick to @mhdawson's proposed schedule, we have time to discuss and converge on this topic.

mhdawson:
@Trott we can go for shorter timeline but as @gireeshpunathil says we should then start in May to target the tenure end of May 30.

mhdawson:
Based on end of term being May 30 and suggestion for 1 week per step I suggest we do:

May 9: Call for Nominations
May 15 : Close nominations
May 16: Open balloting
May 23: Close balloting
May 24: Announce results and begin handover (if any is needed)

We can then repeat the same every May.

@gireeshpunathil maybe you can bring up the suggestion for a vice chair in the TSC meeting this week to get the discussion started. With this proposed schedule we have a three of weeks to discuss before we'd start the process.

mhdawson:
Today was the day we agreed to open nominations and we agreed to vote on a Chair and vice-chair position.  If you would like to run please post in the issue and include which position(s) you would like to run for.

mhdawson:
I'm running for the position of chair again this year.

PR #1219, opened 2022-04-29
doc: add TSC vice chairperson

mhdawson:
Add TSC vice chairperson as discussed in recent
meeting.

Signed-off-by: Michael Dawson <mdawson@devrus.com>

### DIFF
--- a/TSC-Charter.md
+++ b/TSC-Charter.md
@@ -53,10 +53,11 @@ The TSC may, at its discretion, invite any number of non-voting observers to
 participate in the public portion of TSC discussions and meetings.
 
 The TSC shall meet regularly using tools that enable participation by the
-community (e.g. weekly on a Google Hangout On Air, or through any other
+community (e.g. weekly on a Zoom call, or through any other
 appropriate means selected by the TSC). The meeting shall be directed by
-the TSC Chairperson. Responsibility for directing individual meetings may be
-delegated by the TSC Chairperson to any other TSC member. Minutes or an
+the TSC Chairperson or vice Chairperson. Responsibility for directing
+individual meetings may be delegated by the TSC Chairperson or vice
+Chairperson to any other TSC member. Minutes or an
 appropriate recording shall be taken and made available to the community
 through accessible public postings.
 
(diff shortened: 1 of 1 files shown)


mhdawson:
@nodejs/TSC as discussed in the latest meeting and in https://github.com/nodejs/TSC/issues/1201

gireeshpunathil (approved):

TSC-Charter.md
```diff
@@ -53,10 +53,11 @@ The TSC may, at its discretion, invite any number of non-voting observers to
...
 
 The TSC shall meet regularly using tools that enable participation by the
-community (e.g. weekly on a Google Hangout On Air, or through any other
+community (e.g. weekly on a Zoom call, or through any other
 appropriate means selected by the TSC). The meeting shall be directed by
-the TSC Chairperson. Responsibility for directing individual meetings may be
-delegated by the TSC Chairperson to any other TSC member. Minutes or an
+the TSC Chairperson or vice Chairperson. Responsibility for directing
+individual meetings may be delegated by the TSC Chairperson or vice
+Chairperson to any other TSC member. Minutes or an
```
Trott, line 60: I don't think we need this change. It already says "Responsibility for directing individual meetings may be delegated by the TSC Chairperson to any other TSC member." So the TSC can choose to delegate a vice chair, and the chair can choose to delegate to the vice chair for any meeting they can't make. If we add the Vice Chair to the Charter, we're tying our hands that we always need to have a Vice Chair, even if we try this and find out it doesn't work.

Trott (changes requested):
I don't think we need this change and I prefer keeping the Charter as non-specific as possible. If it's permitted by the Charter already (and I think it is), then we don't need to add it to the Charter.

Trott:
> I don't think we need this change and I prefer keeping the Charter as non-specific as possible. If it's permitted by the Charter already (and I think it is), then we don't need to add it to the Charter.

In other words, let's close this and just get right to the business of electing (or selecting) one or more Vice Chairs.

mcollina (approved):
lgtm

mhdawson:
Ok, if we don't need to document the vice-chair anywhere I'm happy to close.

PR #1277, opened 2022-08-31
chore: add votes for re-adding TSC members

Trott:


apapirovski (approved):

Trott:
One thing we could do is add James and Franzi as voters so that they aren't flagged for removal again for another 3 months. However, just to match reality, we can leave it and close any weekly PRs that try to remove them. (Hopefully those will stop pretty soon due to attendance and/or voting. But if it doesn't *and* it gets annoying, we can always add it after-the-fact.)

cjihrig (approved):

aduh95 (approved):

PR #1297, opened 2022-10-21
fix: add missing header in minutes doc

Trott:
Our tooling depends on this header so we should land this quickly before the find-inactive-folks job runs and wrongly flags James again. The heading structure of the minutes is wrong without it anyway. Announcements should not be a sub-header of Present.


BethGriggs (approved):

mhdawson (approved):
LGTM

cjihrig (approved):

PR #1301, opened 2022-10-26
chore: update membership listings in Security-Team.md

Trott:


Trott:
(This will need to be synch'ed again when bmeck accepts the GitHub invitation, but I don't think that's any reason to not approve and land this right away.)

BethGriggs (approved):

Trott:
> (This will need to be synch'ed again when bmeck accepts the GitHub invitation, but I don't think that's any reason to not approve and land this right away.)

Updated with Bradley added.

PR #1350, opened 2023-03-07
Update charter to address membership challenges

Trott:
Change "TSC Member" to "TSC Voting Member", and change "TSC Emeritus Member" to "TSC Member". This lets the TSC somewhat mirror the CPC, where only about 1/3 of the members actually vote (and most of the other members aren't in fact active, but it doesn't adversely affect the CPC).

### DIFF
--- a/TSC-Charter.md
+++ b/TSC-Charter.md
@@ -17,60 +17,47 @@ In the case of the Node.js project, it is delegated to the Node.js
 Technical Steering Committee (“TSC”). OpenJS Foundation’s business
 leadership is the Board of Directors (the “Board”).
 
-This Technical Steering Committee Charter reflects a carefully
-constructed balanced role for the TSC and the CPC in the governance of
-the OpenJS Foundation. The charter amendment process is for the TSC to
-propose changes using simple majority of the full TSC, the proposed
-changes being subject to review and approval by the CPC. The CPC may
-additionally make amendments to the TSC charter at any time, though the
-CPC will not interfere with day-to-day discussions, votes or meetings
-of the TSC.
+The charter amendment process is for the voting members of the TSC to propose
+for review and approval by the CPC. The CPC may additionally make amendments to
+the TSC charter at any time, though the CPC will not interfere with day-to-day
+discussions, votes or meetings of the TSC.
 
 ## Section 3. Establishment of the TSC
 
 TSC memberships are not time-limited. There is no maximum size of the TSC.
-The size is expected to vary in order to ensure adequate coverage of important
-areas of expertise, balanced with the ability to make decisions efficiently.
-The TSC must have at least four members.
+The TSC must have at least four voting members.
 
 There is no specific set of requirements or qualifications for TSC
-membership beyond these rules. The TSC may add additional members to the
+membership beyond these rules. The TSC may add additional voting members to the
 TSC by a standard TSC motion. A TSC member may be removed from the
-TSC by voluntary resignation, by a standard TSC motion, or in accordance to the
-participation rules described below.
+TSC by voluntary resignation or by a standard TSC motion.
 
-Changes to TSC membership should be posted in the agenda, and may be suggested
-as any other agenda item.
-
-No more than one-fourth of the TSC members may be affiliated with the
-same employer. If removal or resignation of a TSC member, or a change of
-employment by a TSC member, creates a situation where more than
-one-fourth of the TSC membership shares an employer, then the situation
+No more than one-fourth of the TSC voting members may be affiliated with the
+same employer. If removal or resignation of a TSC voting member, or a change of
+employment by a TSC voting member, creates a situation where more than
+one-fourth of the TSC voting membership shares an employer, then the situation
 must be immediately remedied by the resignation or removal of one or more
-TSC members affiliated with the over-represented employer(s).
-
-The TSC may, at its discretion, invite any number of non-voting observers to
-participate in the public portion of TSC discussions and meetings.
+TSC voting members affiliated with the over-represented employer(s).
 
 The TSC shall meet regularly using tools that enable participation by the
 community (e.g. weekly on a Google Hangout On Air, or through any other
 appropriate means selected by the TSC). The meeting shall be directed by
 the TSC Chairperson. Responsibility for directing individual meetings may be
-delegated by the TSC Chairperson to any other TSC member. Minutes or an
+delegated by the TSC Chairperson to any other TSC voting member. Minutes or an
 appropriate recording shall be taken and made available to the community
 through accessible public postings.
 
-TSC members are expected to regularly participate in TSC activities.
+TSC voting members are expected to regularly participate in TSC activities.
 
-A TSC member is automatically removed from the TSC if, during a 3-month period,
-all of the following are true:
+A TSC voting member is automatically converted to a TSC regular member if,
+during a 3-month period, all of the following are true:
 
 * They attend fewer than 25% of the regularly scheduled meetings.
-* They do not participate in any TSC votes.
+* They do not participate in three consecutive TSC votes.
 
 ## Section 4. Responsibilities of the TSC
 
-Subject to such policies as may be set by the CPC, the TSC is
+Subject to such policies as may be set by the CPC, the TSC voting members are
 responsible for all technical development within the Node.js project,
 including:
 
@@ -85,20 +72,20 @@ including:
 * Mediating technical conflicts between Collaborators or Foundation
   projects.
 
-The TSC will define Node.js project’s release vehicles.
+The TSC voting members will define Node.js project’s release vehicles.
 
 ## Section 5. Node.js Project Operations
 
-The TSC will establish and maintain a development process for the Node.js
-project. The development process will establish guidelines
+The TSC voting members will establish and maintain a development process for the
+Node.js project. The development process will establish guidelines
 for how the developers and community will operate. It will, for example,
 establish appropriate timelines for TSC review (e.g. agenda items must be
 published at least a certain number of hours in advance of a TSC
 meeting).
 
 The TSC and entire technical community will follow any processes as may
-be specified by the OpenJS Foundation Board relating to the intake and license compliance
-review of contributions, including the OpenJS Foundation IP Policy.
+be specified by the OpenJS Foundation Board relating to the intake and license
+compliance review of contributions, including the OpenJS Foundation IP Policy.
 
 ## Section 6. Elections
 
@@ -117,8 +104,8 @@ election is required if there is only one candidate and no objections to
 the candidate's election. Elections shall be done within the projects by
 the Collaborators active in the project.
 
-The TSC will elect from amongst voting TSC members a TSC Chairperson to
-work on building an agenda for TSC meetings and a OpenJS
+The TSC voting members will elect from amongst voting TSC members a TSC
+Chairperson to work on building an agenda for TSC meetings and a OpenJS
 Cross Project Council (CPC) voting member to represent the TSC in
 the OpenJS Foundation for a term of one year. The Chair and voting CPC
 member may be (but are not required to be) the same person.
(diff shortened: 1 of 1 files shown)


Trott:
@nodejs/tsc

TSC-Charter.md
```diff
@@ -17,60 +17,47 @@ In the case of the Node.js project, it is delegated to the Node.js
...
+TSC voting members are expected to regularly participate in TSC activities.
 
-A TSC member is automatically removed from the TSC if, during a 3-month period,
-all of the following are true:
+A TSC voting member is automatically converted to a TSC regular member if,
+during a 3-month period, all of the following are true:
 
 * They attend fewer than 25% of the regularly scheduled meetings.
-* They do not participate in any TSC votes.
+* They do not participate in three consecutive TSC votes.
```
GeoffreyBooth, line 56: I assume the meetings thing is already pending approval somewhere else? So assuming that goes through, why not just this?

```
A TSC voting member is automatically converted to a TSC regular member if
they do not participate in three consecutive TSC votes.
```

TSC-Charter.md
```diff
@@ -17,60 +17,47 @@ In the case of the Node.js project, it is delegated to the Node.js
...
+the TSC charter at any time, though the CPC will not interfere with day-to-day
+discussions, votes or meetings of the TSC.
 
 ## Section 3. Establishment of the TSC
 
 TSC memberships are not time-limited. There is no maximum size of the TSC.
-The size is expected to vary in order to ensure adequate coverage of important
-areas of expertise, balanced with the ability to make decisions efficiently.
-The TSC must have at least four members.
+The TSC must have at least four voting members.
```
apapirovski, line 28: I think you might need to define voting and non-voting members somewhere? Right now there's some amount left unsaid.

Presumably, the new rules imply that one has to be added to the TSC as a voting member and can only transition to non-voting membership by failing to meet the participation criteria?

Similarly, the "removal" aspect becomes a bit less clear. The wording right now still requires resignation/removal of TSC voting members due to employee affiliation quotas but wouldn't being moved to a regular member be enough?

TSC-Charter.md
apapirovski, line 56: This is sort of more nit-picky but I'm having trouble following whether this 3-month removal only applies if there are 3 votes within a 3 month period or if it's "attend fewer than 25% of regularly scheduled meetings in a 3-month period AND do not participate in three consecutive TSC votes"?

Trott: Yes, I think you're right.

TSC-Charter.md
Trott: Yes, room for improvement/clarity there for sure. (I did this very quickly because I suspected that if I didn't do it in a 15-minute sprint, I probably wouldn't come back to it for a long time.)

TSC-Charter.md
Trott: I think we should remove the 3-month part entirely as well as the meeting requirement (as previously discussed and as Geoffrey points out above), and just have three consecutive votes be it. I suppose we could add some text to protect against three votes happening in one day or something like that, but I'm personally not too worried about that. (Votes are usually open for a week or two anyway, or at least a few days. And they are infrequently concurrent.)

mcollina: I would change it so votes are taken asynchronously (not during meetings) and the pools are open for at least a week.

Trott: Done.

Trott: > This is sort of more nit-picky but I'm having trouble following whether this 3-month removal only applies if there are 3 votes within a 3 month period or if it's "attend fewer than 25% of regularly scheduled meetings in a 3-month period AND do not participate in three consecutive TSC votes (in any given time-frame)"?

I think this is resolved with the change pushed just now, but if I'm wrong, please leave another comment.

TSC-Charter.md
Trott: I've added a paragraph at the start of section 3 that explains what regular members and voting members are. I'm going to optimistically mark this as resolved, but feel free to unresolve it or leave an additional comment if you think more should be done.

aduh95 (commented):
If we make this change, we need first to discuss the org admin aspect of being a TSC member, if we have TSC members who are inactive they don’t need to be org admins (not sure about HackerOne, and the TSC mailing list, worth discussing as well), and keeping their access is just added risk for Node.js to have one of its admin account compromised. The easiest solution would be to restrict this to TSC voting members; another solution would be to have another subset of the TSC that would be appointed as org admin for e.g. one year.

Trott:
> If we make this change, we need first to discuss the org admin aspect of being a TSC member,

My intention was that only voting members would be org admins, HackerOne members, email recipients (although we can have two separate TSC mailing lists--one for all members and one for voting members), etc. You are right that we'll want to figure that all out ahead of time. (And don't forget the TSC private Slack channel!) I think everything (except for maybe the TSC Slack channel) is probably covered in the TSC onboarding/offboarding docs so that's a good place to look to see if we are forgetting about anything.

GeoffreyBooth:
> I want to get an idea if there is support or not for the general concept

Yes, I support the general concept. We might want to use this as an opportunity to reconsider the _name_ of the TSC, as we’re really the executive committee or leadership committee of the project and responsible for more than just “technical steering.” Regardless of the name of “_ Committee / _ Committee Voting Member” I think the proposed structure will be a significant improvement.

RafaelGSS (approved):

Trott:
@tobie @eemeli You two had especially good/thoughtful comments on the last proposed charter change. Any chance you can take a look at this before we pass it on to the CPC voting members for approval?

mcollina (approved):
lgtm

TSC-Charter.md
```diff
@@ -129,41 +117,37 @@ of terms a TSC Chairperson or voting CPC member may serve.
...
 periods) within the development process.
 
-The TSC follows a [Consensus Seeking][] decision making model. When an agenda
-item has appeared to reach a consensus the moderator will ask "Does anyone
-object?" as a final call for dissent from the consensus.
-
-If an agenda item cannot reach a consensus a TSC member can call for
-either a closing vote or a vote to table the issue to the next meeting.
-The call for a vote must be seconded by a majority of the TSC or else the
-discussion will continue.
```
eemeli, line 143: Why is this dropped as a part of this PR? Is it to remove this tabling "vote" from the count for voting/regular member status?

eemeli (approved):
This looks much better!

TSC-Charter.md
```diff
@@ -17,60 +17,48 @@ In the case of the Node.js project, it is delegated to the Node.js
...
-CPC will not interfere with day-to-day discussions, votes or meetings
-of the TSC.
+This charter can only be amended with the approval of the CPC.
 
 ## Section 3. Establishment of the TSC
 
+TSC members can be either _regular_ members or _voting_ members. Regular
+members can attend meetings and participate in TSC discussions, but do not
+vote. Voting members can do everything regular members can do, and also have
+the ability to cast votes when consensus is not reached on an issue.
```
jasnell, line 27: eh, I'm really not all that convinced this is worthwhile, to be honest. Yes, I understand we've had some issues here but I think we need to do a better job of encouraging no-longer active TSC members to just not be TSC members anymore. I'm not going to block, however, if the majority of TSC members want this. Just consider my vote to be -0.

TSC-Charter.md
Trott: I removed it because I have been on the TSC for I-don't-know-how-many years and this process has never once been used. So I wanted to take this opportunity to simplify the charter so that it reflects our actual processes. However, I can put it back in and remove it in a separate proposal if you prefer.

nschonni:
I'm having a hard time seeing the difference between a non-voting TSC member and the existing Collaborator role. Possibly the only one might be that they'd be invited to in-camera TSC meetings, while Collaborators wouldn't?

Trott:
> I'm having a hard time seeing the difference between a non-voting TSC member and the existing Collaborator role. Possibly the only one might be that they'd be invited to in-camera TSC meetings, while Collaborators wouldn't?

Yes, the differences are small and possibly non-existent. The charter as proposed here doesn't require it, but I imagine that non-voting members would be invited to the meetings and also that we might choose to create one or more communication channels specifically for TSC members (both voting and non-voting) such as an email list or a Slack channel. But strictly speaking, we might do neither of those things and it is just a title in that case, with no additional privileges or responsibilities.

TSC-Charter.md
eemeli: I have no preference on this, really just wanted an explicit comment/explanation like this available somewhere.

tobie:
That's a great improvement on the previous proposal, and simplifies things quite a bit.

I do agree with @jasnell, however that the distinction between the two kinds of members does still feel like a lot of governance in order to avoid just _asking_ people to step down. Have you tried for the chair to reach out to folks that stop showing up in order to understand what's going on, help address possible issues, and suggest stepping down otherwise, or being on "pause" for a while?

I've also heard mentioned that reaching quorum could be an issue. If so, have y'all considered calculating quorum against the set of folks who had voted in the last three votes (for example) rather than against the whole membership?

Overall, I understand that this is a sensitive topic and that spelling out what the underlying issues is might be unpleasant, but I'm also wary of coming up with solutions to problems that are not clearly specified.

That said, this is still a great improvement of the previous proposal and the existing solution, and as I said from the get-go, the role of the CPC isn't to dictate how projects run, but to make sure that they do so in a way that is compatible with the values of the OpenJSF (which all of those different propositions _absolutely_ are), and to offer suggestions/guidance when prompted. My bias is towards simplifying things as much as possible and getting to the desired outcomes through communication, culture, and values rather than policy, so I'd encourage you to explore that direction further. But this is also good enough, and moving ahead with this (and revisiting down the line if necessary)is also a completely sound approach.

tobie:
> We can do that in _addition_ to these charter changes, but I would not want to do that _instead_.

That sounds reasonable.

> And to preempt a potential question: "Won't that make the title of 'TSC member' less valuable or even worthless?" My answer is that would be a feature and not a bug. If this change makes TSC (voting) member less of a symbol of status or power to be pursued, and more of a duty, volunteer function, or perhaps even a calling, that would be good.

Can't agree more.

tniessen:
We would also have to update docs within the main repository based on this change. For example, if TSC emeriti become "TSC members", my understanding is that their approvals should still not count as TSC approvals on semver-major PRs.

RaisinTen (approved):

Trott:
> We would also have to update docs within the main repository based on this change. For example, if TSC emeriti become "TSC members", my understanding is that their approvals should still not count as TSC approvals on semver-major PRs.

Yes. My thinking is:

1. Get TSC consensus on this change.
2. Get CPC approval for this change.
3. Update all the docs in the main repo.
4. Land this PR.

I suppose there's nothing stopping me or someone else from pulling together a PR for step 3 right now, as it seems likelier-than-not that this has consensus. On the other hand, maybe I (or someone else) should wait until there are more than 4 current TSC members (including me) indicating they support this change.

apapirovski (approved):

GeoffreyBooth (approved):

danielleadams (approved):

targos (approved):

joesepi (approved):

legendecas (approved):

mhdawson:
I'm -0 on this as well.

Trott:
> I'm -0 on this as well.

Why?

TSC-Charter.md
Trott: I believe https://github.com/nodejs/TSC/pull/1350#issuecomment-1464091435 addresses this line of thinking.

cjihrig:
I'm also a 0 on this.

Trott:
CPC approval has been attained. The lazy consensus here (9 TSC approvals, 3 TSC members taking the no-objection-but-not-approving-either position, open for a week, discussed at meetings and in Slack) means this is good to land.

I'll open a PR to land in the main repo to change the wording in documentation there so it can land at the same time as this.

aduh95 (approved):
I would prefer to let the TSC as an entity the subject of most of these sentences, it might be technically equivalent, however I would much prefer putting the emphasis on the group rather than the individuals that constitute it. What I would like to understand when reading this charter is that being a (voting) member of the TSC does not grant any power over other collaborators (well there’s the org admin part of it, but I cannot override an objection from a collaborator), only the TSC has specific powers. But maybe my review’s coming a bit too late, so feel free to ignore.

TSC-Charter.md
```diff
@@ -17,60 +17,48 @@ In the case of the Node.js project, it is delegated to the Node.js
...
 
-* They attend fewer than 25% of the regularly scheduled meetings.
-* They do not participate in any TSC votes.
+A TSC voting member is automatically converted to a TSC regular member if
+they do not participate in three consecutive TSC votes.
 
 ## Section 4. Responsibilities of the TSC
 
-Subject to such policies as may be set by the CPC, the TSC is
+Subject to such policies as may be set by the CPC, the TSC voting members are
```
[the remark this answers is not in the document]
GeoffreyBooth: I agree. Isn’t the process for any non-voting TSC member to become a voting TSC member to simply vote, or ask to become a voting member? So de facto the entire TSC is responsible for everything, because anyone can choose to “step forward” into voting status at any time. So we can just say “the TSC” for most things.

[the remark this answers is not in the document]
aduh95: Regular members can't vote: "Regular members can attend meetings and participate in TSC discussions, but do not vote.". The process to become a voting member for a regular member is outlined in the document: "A standard TSC motion can be used to change a regular TSC member to a voting TSC member, or to change a voting TSC member to a regular TSC member." And I think a standard TSC motion means lazy consensus over the whole TSC, and a vote (among the voting members) if no consensus can be reached? But if regular members are mostly inactive, does that mean we will be forced to move to a vote, or can we make a motion without the majority of the TSC participating?

bnb:
Commenting as I'm reading, but wanted to get ideas out:

This is cool and I'm happy to see this change. Overall I think this will be a healthy change for the project. IMO this will dramatically reduce the barrier to people feeling like they belong, which I think is likely a good move for the project and how it's rather uniquely governed.

> I've done that kind of outreach in the past and it has never yielded results. People always want to stay on the TSC, typically because "Node.js TSC member" helps people's careers and conference-speaking prospects and other things. By allowing them to keep the title, this change will (hopefully) make the process easier. (I have already had one person tell me they would probably move to the "regular member" status if this change goes through, which is certainly an indication that this change might make the whole role-change less challenging for everyone.)

I'll 100% +1 this - IMO this is a result of the extremely relaxed membership approach that io.js took, which defined the culture of participation in Node.js to this day. I had a very similar experience as the CommComm chair - people would just say yes they wanted to be a member then never show up again. Every time I asked they were just about to be able to contribute again. I've definitely ended up in this position myself a few times - though in those situations in the past year or two, I've tried to recognize it and step back to evaluate if my response is the most accurate one I can give. 

Adding Emeritus with a low barrier in CommComm did help with this, and I'm guessing the model proposed here is going to be even more helpful.

> And to preempt a potential question: "Won't that make the title of 'TSC member' less valuable or even worthless?" My answer is that would be a feature and not a bug. If this change makes TSC (voting) member less of a symbol of status or power to be pursued, and more of a duty, volunteer function, or perhaps even a calling, that would be good.

This is a good take and I'm happy to hear it.

In general, I have been worried about stagnation in the project's leadership. We used to see a lot more faces cycling through higher leadership positions, but that has dramatically slowed. I'm hopeful that this is a healthy step that will enable more people grow into leadership in the project.

TSC-Charter.md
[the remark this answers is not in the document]
GeoffreyBooth: Is there a reason to require a motion for a non-voting member to become active again? I would think that we would want as low a bar (if any) as possible, so that no one feels any need to fight to stay as a voting member when they’ve missed votes and are automatically recategorized.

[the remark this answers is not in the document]
aduh95: With all the things tied to being a voting member (e.g. being an org admin), I’d say we do want a not too low bar.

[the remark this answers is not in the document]
gireeshpunathil: how about this clause instead of TSC motion for re-entry: `A TSC regular member is automatically converted to a TSC voting member if they participate in three consecutive TSC meetings`?

 - reasonable re-entry criteria that balances with exit criteria
 - meets TCS member's band-width / responsibility tradeoff
 - everyone knows the rules and easy to follow
 - scripts are easy to develop and run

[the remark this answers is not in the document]
aduh95: Note that this proposal doesn't take into account folks who might want to be TSC members without being  voting members (but maybe no one wants that, I don't know).
Personally, I'm -1 on any automation that adds folks as org admin without at least a TSC motion. If we can come up with a proposal that dissociates "being a voting TSC member" and "being an org admin", I would love that but that seems easier said than done.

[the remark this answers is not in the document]
Trott: IMO, it is too late to change the text on this PR. It has been approved by the CPC and previously approved by the TSC. Coming in now to suggest changes is too late. It is already our charter. It just hasn't landed because a few other PRs need to land too to keep everything consistent.

If people want to change the charter further, that would be a new PR.

Trott:
Landing as-is, as this text was approved by CPC and TSC, and comments remaining are listed as non-blocking.  Feel free to open another PR if we want to revise the text further and thank you, everyone, for your patience with all of this.

PR #1360, opened 2023-03-22
Update charter to concentrate responsibilities on the TSC rather than its voting members

aduh95:
~The only effect of this change should be that TSC regular members participate in lazy consensus process~ EDIT: this has been removed, this PR should not have any effect for regular members. If you see another effect that I overlooked, that's not on purpose, so please chime in.

The aim of this change is to clarify that TSC voting members do not have any specific powers (except the ones defined by the Moderation Policy and the ones delegated by the TSC on a case by case basis) over Collaborators.

### DIFF
--- a/TSC-Charter.md
+++ b/TSC-Charter.md
@@ -121,7 +121,7 @@ Consensus. The TSC shall establish appropriate guidelines for
 implementing Lazy Consensus (e.g. expected notification and review time
 periods) within the development process.
 
-The TSC follows a [Consensus Seeking][] decision making model.
+The TSC follows a [Consensus Seeking][] decision making model among voting members.
 When an agenda item has appeared to reach a consensus the moderator will ask
 "Does anyone object?" as a final call for dissent from the consensus.
 
--- a/TSC-Charter.md
+++ b/TSC-Charter.md
@@ -145,11 +145,10 @@ basis as defined in the development process.
 Collaborators may opt to elevate significant or controversial
 modifications, or modifications that have not found consensus to the TSC
 for discussion by assigning the `tsc-agenda` tag to a pull request or
-issue. When consensus cannot be reached, the TSC voting members should serve as the
-final arbiter by casting a vote.
-The TSC will maintain and publish a list of current Collaborators, as
-well as a development process guide for Collaborators and Contributors
-looking to participate in the development effort.
+issue. When consensus cannot be reached, the TSC voting members should serve as
+the final arbiter by casting a vote. The TSC will maintain and publish a list of
+current Collaborators, as well as a development process guide for Collaborators
+and Contributors looking to participate in the development effort.
 
 ## Section 9. Definitions
 
--- a/TSC-Charter.md
+++ b/TSC-Charter.md
@@ -58,7 +58,7 @@ they do not participate in three consecutive TSC votes.
 
 ## Section 4. Responsibilities of the TSC
 
-Subject to such policies as may be set by the CPC, the TSC voting members are
+Subject to such policies as may be set by the CPC, the TSC is
 responsible for all technical development within the Node.js project,
 including:
 
@@ -73,11 +73,11 @@ including:
 * Mediating technical conflicts between Collaborators or Foundation
   projects.
 
-The TSC voting members will define Node.js project’s release vehicles.
+The TSC will define Node.js project’s release vehicles.
 
 ## Section 5. Node.js Project Operations
 
-The TSC voting members will establish and maintain a development process for the
+The TSC will establish and maintain a development process for the
 Node.js project. The development process will establish guidelines
 for how the developers and community will operate. It will, for example,
 establish appropriate timelines for TSC review (e.g. agenda items must be
@@ -117,11 +117,11 @@ of terms a TSC Chairperson or voting CPC member may serve.
 ## Section 7. Voting
 
 For internal project decisions, Collaborators shall operate under Lazy
-Consensus. The TSC voting members shall establish appropriate guidelines for
+Consensus. The TSC shall establish appropriate guidelines for
 implementing Lazy Consensus (e.g. expected notification and review time
 periods) within the development process.
 
-The TSC voting members follow a [Consensus Seeking][] decision making model.
+The TSC follows a [Consensus Seeking][] decision making model.
 When an agenda item has appeared to reach a consensus the moderator will ask
 "Does anyone object?" as a final call for dissent from the consensus.
 
@@ -134,8 +134,7 @@ All changes to this charter must be approved by the CPC.
 ## Section 8. Project Roles
 
 The Node.js git repository is maintained by the TSC and
-additional Collaborators who are added by the TSC voting members on an ongoing
-basis.
+additional Collaborators who are added by the TSC on an ongoing basis.
 
 Individuals making significant and valuable contributions are made Collaborators
 and given commit-access to the project. These individuals are identified by the
@@ -146,7 +145,7 @@ basis as defined in the development process.
 Collaborators may opt to elevate significant or controversial
 modifications, or modifications that have not found consensus to the TSC
 for discussion by assigning the `tsc-agenda` tag to a pull request or
-issue. The TSC voting members should serve as the final arbiter where required.
+issue. TSC votes should serve as the final arbiter where required.
 The TSC will maintain and publish a list of current Collaborators, as
 well as a development process guide for Collaborators and Contributors
 looking to participate in the development effort.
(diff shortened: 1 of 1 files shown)


TSC-Charter.md
```diff
@@ -146,7 +145,7 @@ basis as defined in the development process.
 Collaborators may opt to elevate significant or controversial
 modifications, or modifications that have not found consensus to the TSC
 for discussion by assigning the `tsc-agenda` tag to a pull request or
-issue. The TSC voting members should serve as the final arbiter where required.
+issue. TSC votes should serve as the final arbiter where required.
```
mcollina, line 148: I think this paragraph should be left as-is.

mhdawson: +1

aduh95: Would the following work better for everyone?
```suggestion
issue. The TSC should serve as the final arbiter where required.
```

mcollina: I don't think we are in agreement with the change. The TSC voting members _are_ the final arbiter.

gireeshpunathil, line 148:
```suggestion
issue. TSC should serve as the final arbiter where required, and specifically it's voting members, when consensus cannot be reached.
```

ljharb: (its)

Trott: I think leaving it as "TSC voting members" removes ambiguity. I think characterizing it as the TSC as a whole is friendly, but not precise. I'd favor precision in this case.

aduh95: Maybe it's because I don't interpret English as you'all, but to me it adds some ambiguity: it could mean that I, as a TSC voting member, can serve as the final arbiter. But if we use "the TSC", it's very clear that we talk about the collective, not the individuals.
Would Gireesh suggestion (https://github.com/nodejs/TSC/pull/1360#discussion_r1145729795) work for everyone? Or does someone else has a suggestion? Or should I just shut up about it already 😅?

mcollina: Ah, I understand where you are coming from @aduh95. While English is not my native language either, I think `the TSC` is not precise.

I think something like the following could work:

> When consensus cannot be reached, the TSC voting members should serve as the final arbiter by casting a vote.

Wdyt?

aduh95: I love it, and I agree your suggestion in indeed less ambiguous.
```suggestion
issue. When consensus cannot be reached, the TSC voting members should serve as the
final arbiter by casting a vote.
```

TSC-Charter.md
```diff
@@ -146,7 +145,8 @@ basis as defined in the development process.
 Collaborators may opt to elevate significant or controversial
 modifications, or modifications that have not found consensus to the TSC
 for discussion by assigning the `tsc-agenda` tag to a pull request or
-issue. The TSC voting members should serve as the final arbiter where required.
+issue. When consensus cannot be reached, the TSC voting members should serve as the
+final arbiter by casting a vote.
 The TSC will maintain and publish a list of current Collaborators, as
 well as a development process guide for Collaborators and Contributors
 looking to participate in the development effort.
```
aduh95, line 152:
```suggestion
issue. When consensus cannot be reached, the TSC voting members should serve as
the final arbiter by casting a vote. The TSC will maintain and publish a list of
current Collaborators, as well as a development process guide for Collaborators
and Contributors looking to participate in the development effort.
```

Trott:
Just a reminder, in case someone is inclined to land this, that this cannot land without CPC review and approval. All changes to the charter need to be approved by the CPC. Once there's consensus, I (or someone else) can take it to the CPC.

gireeshpunathil (approved):

mcollina (approved):
lgtm

Trott:
It's a minor thing, but I would prefer the commit message and this PR title to refer to "responsibilities" (or something similar) rather than "powers". I would prefer language that focuses on the work and/or its impact, and not on power and/or status.

Trott:
Given how old this is and that it has not (to my knowledge) been brought to the OpenJS CPC, and that there is no intention of bringing it to the CPC (to my knowledge, at least), perhaps this should be closed?

aduh95:
Why is there no intention of bringing it to the CPC? I'd like it to be.

Trott:
> Why is there no intention of bringing it to the CPC? I'd like it to be.

Ah, well, Joe Sepi and Matteo Colina are the Node.js voting CPC members, so if there is TSC consensus on this, they're probably the best candidates to put it on the CPC agenda.

Personally, I don't fell like we should do this. It is well-intentioned, and clarifying limits to TSC voting members is fine, but this seems to elevate TSC regular members. But for better or for worse, the whole point of "regular TSC members" is to allow people to retain the TSC status without retaining authority and responsibility. (I know that's the whole point because I'm the one that proposed the status. It's also why the term "regular member" is not exactly intuitive, unfortunately.) Prior to the creation of the "regular member" status, we had people (in my opinion) game the system to retain their TSC role without actually being invested in the project. This (in my opinion) created and exacerbated all sorts of problems. Creating "regular member" status allowed people to not freak out about how being removed from active TSC status might affect their career prospects or whatever. So they could let go of being a TSC member (because they were technically still TSC members, even if they don't actually do anything any other collaborator wouldn't/couldn't do).

The change proposed here kind of undoes that. Except for maybe Geoffrey Booth, none of the current Regular Members (including me!) are (as far as I know) sufficiently engaged that they should be considered stewards of the project to a significantly greater degree than other active collaborators. I don't see the advantage of granting them responsibility or authority beyond that of collaborators (other than that it makes "regular member" more intuitive, which I get, but unfortunately I think the surprising meaning of "regular member" is a feature, not a bug).

All of that said, I consider this a decision for the voting members (like you!) to make and not the regular members (like me). So...just my opinion, and quite possibly I'm the only one that holds that opinion. (I don't feel great about it, particularly the way it makes "regular member" a misleading term, but we tried a few different approaches to addressing inactive TSC membership, and this was the only one that worked in my opinion.)

<sub>Since I know someone is going to say some variation on "we can create measurable criteria and then automate removal of people from the TSC": We tried that. That's what I mean by "game the system" above. If people know they need to do to keep membership, they will do the minimum but it will not result in them being engaged.</sub>

TSC-Charter.md
```diff
@@ -117,11 +117,11 @@ of terms a TSC Chairperson or voting CPC member may serve.
 ## Section 7. Voting
 
 For internal project decisions, Collaborators shall operate under Lazy
-Consensus. The TSC voting members shall establish appropriate guidelines for
+Consensus. The TSC shall establish appropriate guidelines for
 implementing Lazy Consensus (e.g. expected notification and review time
 periods) within the development process.
 
-The TSC voting members follow a [Consensus Seeking][] decision making model.
+The TSC follows a [Consensus Seeking][] decision making model.
```
aduh95, line 124:
```suggestion
The TSC follows a [Consensus Seeking][] decision making model among voting members.
```

aduh95:
> this seems to elevate TSC regular members

It's not the intention of this PR, the goal is to rephrase the charter so individual voting members do not hold power, only the TSC as a group does, I'm hopeful that it can be achieved without changing the status of regular members. If you have in mind some specific change that's problematic, please point it out, and let's try to find a better way for phrasing it.

Trott:
> the goal is to rephrase the charter so individual voting members do not hold power, only the TSC as a group does

Ah! That makes a lot of sense. Thanks for explaining.

mhdawson:
The challenge I seen in having regular members participate in lazy concensus is that what happens when we discuss private issues and need to get consensus. Regular members are not on any of the existing mailing lists and are not currenty involved in private discussions.  In some cases we even use email and or slack to be extra sure all TSC members have seen an issue so technical it might apply to non private issues as well. For the reasons @Trott mentioned I'd prefer not to create more work/process for what is a effectively a cerimonial title.

aduh95:
@mhdawson that's fair, that's what I've tried to address with https://github.com/nodejs/TSC/pull/1360/commits/6ac4115c7aa53eea279b00da5f1c7164b0515600, let me know if that works for you

mhdawson (approved):
LGTM

mhdawson:
@aduh95 thanks for updating, added my LGTM.

mhdawson:
@aduh95 was this replaced by https://github.com/nodejs/TSC/pull/1636 which just landed or was it a follow on?

aduh95:
> @aduh95 was this replaced by #1636 which just landed or was it a follow on?

It is two different things, https://github.com/nodejs/TSC/pull/1636 deals with the way the TSC votes (i.e. technical), this one is trying to clarify that TSC is holding the "powers" defined by the charter, not its individual members.

jasnell (approved):

mcollina (approved):
lgtm

tobie:
**Notes from this week's CPC call:** CPC approved this charter change

PR #1374, opened 2023-04-19
Add sharing TSC meeting password to onboarding

richardlau:


richardlau:
I'm not sure if this also needs to be in the section for regular (non-voting) members, or in a new section common to both.

mhdawson (approved):
LGTM

tniessen (approved):

aduh95:
> I'm not sure if this also needs to be in the section for regular (non-voting) members, or in a new section common to both.

I think regular members also need to have that password as they are also invited to meetings IIUC.

tniessen:
> I'm not sure if this also needs to be in the section for regular (non-voting) members, or in a new section common to both.

AFAIK the only way to become a regular members is by being a voting member first, in which case the member already has the password.

ronag (approved):

MoLow (approved):

gireeshpunathil (approved):

RaisinTen (approved):

Trott (approved):

PR #1496, opened 2024-01-24
doc: add minutes for meeting Jan 24 2024

mhdawson:


meetings/2024-01-24.md
```diff
@@ -0,0 +1,116 @@
...
+    * yarn and pnpm have large enough downloads,
+    * npm don’t want to support/integrate with corepack as it does not fit with view of how
+      package managers should be integrated
+    * corepack was started due to context that limited willingness to add yarn/pnpm to Node.js,
+      but extended beyond that.
+    * believe we should make yarn and pnpm just work, but that does not necessarily mean
+      corepack
+    * should define governance for supporting clients
+  * Michael
+  * Need to factor in work on the project to support any particular solution
```
aduh95, line 50:
```suggestion
  * Michael
    * Need to factor in work on the project to support any particular solution
```
or
```suggestion
  * Michael: Need to factor in work on the project to support any particular solution
```

meetings/2024-01-24.md
```diff
@@ -0,0 +1,116 @@
...
+  * Myles
+    * Question, is are we going to ship these and take on responsibility for the clients
+    * Corepack tries to take those on, but does it actually let us say “no” to us taking on
+      responsibility
+  * James, do believe it’s an important distinction that it does not bundle the clients that should
+    not be lost. Exploring different options does make sense.
+  * Myles yarn, pnpm just working can lead people to thinking its the same, which is a concern.
+  * Myles in support of finding equitable environment
+  * Richard - installing things, having up front questions, that is how npx works, current
+    corepack, seems similar,  current npx asks if you are going to install
```
aduh95, line 68:
```suggestion
  * Richard - when installing things, having up front questions; that is how npx works;
    corepack like npx downloads software that we are not responsible for.
```

aduh95 (approved):

gireeshpunathil (approved):

MoLow (approved):

meetings/2024-01-24.md
```diff
@@ -0,0 +1,116 @@
...
+    * npm don’t want to support/integrate with corepack as it does not fit with view of how
+      package managers should be integrated
+    * corepack was started due to context that limited willingness to add yarn/pnpm to Node.js,
+      but extended beyond that.
+    * believe we should make yarn and pnpm just work, but that does not necessarily mean
+      corepack
+    * should define governance for supporting clients
+  * Michael
+    * Need to factor in work on the project to support any particular solution
+  * Richard, if npm was not currently bundled we probably would not bundle, so if we define
```
benjamingr, line 51: @richardlau @MylesBorins @mhdawson I was unable to attend the meeting but the part I don't understand is with regards to @mcollina 's point from last week.

Currently there is exactly one package manager hosting a public registry (with a few mirrors), all other package managers are npm clients that talk to the npm registry. npm can (but doesn't) break other clients but that's out off kindness.

The situation isn't equitable because exactly one client is footing the server bill, not because one client is bundled - and I suspect that had we made the choice today npm would still be bundled or we would ship our own registry (which would incur high costs and would require the project to foot the bill instead of fGitHub).

aduh95: There's obviously more than one public registry, npm is "just" the most popular one especially for the JS ecosystem. Let's not start discussing npm intents, whether it's kindness or whatever, we don't know that (unless you do?), and I don't think it's relevant. FWIW I agree with Richard, if npm wasn't bundled today, we would probably not bundle it; my guess is that npm themselves would distribute `node` – and that would probably work better for everyone.

(You probably already know that, this PR is about keeping notes of what folks say, so this discussion should not block the PR from landing)

richardlau: My point was more that if we were to define criteria today for bundling a package manager with Node.js it is likely that npm wouldn't meet that criteria (see for example, historic tension around npm's lack of long-term support) (FTR I have no idea if any of the other package managers would meet such criteria either). Personally I do not consider that a reason by itself to suggest not bundling npm now as it would be too disruptive to the ecosystem.

richardlau (approved):

meetings/2024-01-24.md
mcollina: Agreed that npm has a release management that is at odds with the one with follow, and it would likely not meet that bar. Specifically around the fact that they use September as the month to ship their major, because we go LTS in October.

@benjamingr can you refer to what I said last week that was not clear?

mhdawson: My 2 cents of clarifying what @mcollina said. It is that because the npm client comes from and is supported by GitHub who hosts the registry, it will always have the distinction of being the "official" client in terms of the provider of the registry (GitHub). Nothing that the node.js project does will change the benefits that distinction brings to the npm client.  For example if something in the registry does not work with the npm client, then GitHub should work to fix that either in the registry or the client. If something does not work with some of the other package managers then they might consider it a problem in the client and not be interested in making changes in the registry even if that means it will be a lot of work for the client to fix the issue.

mcollina: Thanks @mhdawson, exactly that.

PR #1536, opened 2024-04-23
Request for vote to reduce inactive collaborator duration

anonrig:
Requesting vote due to the blockers of pull-request https://github.com/nodejs/node/pull/52459.

Currently, there are 4 blockers (2 TSC members) and 6 approvals (5 TSC members) in the original pull-request.

cc @nodejs/tsc

### DIFF
--- a/votes/initiateNewVote/_EDIT_ME.yml
+++ b/votes/initiateNewVote/_EDIT_ME.yml
@@ -46,7 +46,7 @@ prBody: |
     - a list of all the commit authors.
     - a list of all the collaborators who are listed as reviewers in the commit message (i.e. folks who have approved the PR). 
     
-  The bot will open a PR to move a collaborator to emeritus if they don’t appear in either of those lists. 
+  The bot will open a PR to move a collaborator to emeritus if they don't appear in either of those lists. 
   This vote is to decide whether we want to reduce the timespan from 12 months to
   9 months.
 
--- a/votes/initiateNewVote/_EDIT_ME.yml
+++ b/votes/initiateNewVote/_EDIT_ME.yml
@@ -24,8 +24,8 @@ headerInstructions: |
 #    voters express their preference for each candidates, no matter how many
 #    there are.
 candidates:
-  - Status Quo: Keep the current in-active collaborator duration of 12 months
-  - Reduce the in-active collaborator duration to 9 months
+  - Status quo: Keep the current inactive collaborator duration of 12 months
+  - Reduce the inactive collaborator duration to 9 months
 
 # 4. Pass the following to false if it's important to keep the candidates in the
 #    order you define above. Presenting candidates in a fixed order tends to
--- a/votes/initiateNewVote/_EDIT_ME.yml
+++ b/votes/initiateNewVote/_EDIT_ME.yml
@@ -41,12 +41,14 @@ footerInstructions: |
 
 # 6. Optionally, insert a brief introduction for the vote PR, in the markdown format.
 prBody: |
-  Currently, Node.js collaborators are considered inactive if they have not
-  contributed to the project for 12 months. The definition of contribution in
-  the context of Node.js collaborators is any approval or review of a pull
-  request, or any commit to the Node.js repository. The proposed reduction
-  in the duration of inactivity for Node.js collaborators, from 12 months
-  to 9 months.
+  Currently, Node.js collaborators are considered inactive if in the
+  past 12 months, they haven't done any of the following:
+  
+  - Authoring any landed commit
+  - Approving any landed commit
+  
+  This vote is to decide whether we want to reduce the timespan from 12 months to
+  9 months.
 
 # 7. Optionally, choose an id that will be used for the branch name as well as
 #    the vote folder name. If not supplied, a UUID will be used.
--- a/votes/initiateNewVote/_EDIT_ME.yml
+++ b/votes/initiateNewVote/_EDIT_ME.yml
@@ -3,7 +3,7 @@
 
 # 1. Select a subject for the vote. This can be a question addressed to the TSC
 #    voting members.
-subject: Reduce in-active collaborator duration
+subject: Reduce inactive collaborator duration
 
 # 2. You can leave the header instructions as is, or modify them if you see fit.
 headerInstructions: |
--- a/votes/initiateNewVote/_EDIT_ME.yml
+++ b/votes/initiateNewVote/_EDIT_ME.yml
@@ -46,7 +46,7 @@ prBody: |
     - a list of all the commit authors.
     - a list of all the collaborators who are listed as reviewers in the commit message (i.e. folks who have approved the PR). 
     
-    The bot will open a PR to move a collaborator to emeritus if they don’t appear in either of those lists. 
+  The bot will open a PR to move a collaborator to emeritus if they don’t appear in either of those lists. 
   This vote is to decide whether we want to reduce the timespan from 12 months to
   9 months.
 
--- a/votes/initiateNewVote/_EDIT_ME.yml
+++ b/votes/initiateNewVote/_EDIT_ME.yml
@@ -41,7 +41,12 @@ footerInstructions: |
 
 # 6. Optionally, insert a brief introduction for the vote PR, in the markdown format.
 prBody: |
-  Currently, Node.js collaborators are considered inactive if in the
+  Currently, the bot looks through all commits on `main` that have landed in the last 12 months, and makes two lists:
+  
+    - a list of all the commit authors.
+    - a list of all the collaborators who are listed as reviewers in the commit message (i.e. folks who have approved the PR). 
+    
+    The bot will open a PR to move a collaborator to emeritus if they don’t appear in either of those lists. 
   past 12 months, they haven't done any of the following:
   
   - Authoring any landed commit
--- a/votes/initiateNewVote/_EDIT_ME.yml
+++ b/votes/initiateNewVote/_EDIT_ME.yml
@@ -3,7 +3,7 @@
 
 # 1. Select a subject for the vote. This can be a question addressed to the TSC
 #    voting members.
-subject: REPLACEME
+subject: Reduce in-active collaborator duration
 
 # 2. You can leave the header instructions as is, or modify them if you see fit.
 headerInstructions: |
@@ -24,8 +24,8 @@ headerInstructions: |
 #    voters express their preference for each candidates, no matter how many
 #    there are.
 candidates:
-  - TODO
-  - TODO
+  - Status Quo: Keep the current in-active collaborator duration of 12 months
+  - Reduce the in-active collaborator duration to 9 months
 
 # 4. Pass the following to false if it's important to keep the candidates in the
 #    order you define above. Presenting candidates in a fixed order tends to
@@ -35,10 +35,18 @@ canShuffleCandidates: true
 # 5. Insert here a short description of the vote objectives and link to the
 #    issue it was discussed on to give the full context.
 footerInstructions: |
-  TBD
+  The objective of this vote is to decide whether duration to determine
+  inactivity of a from 12 months to 9 months. This vote was discussed
+  in pull-request https://github.com/nodejs/node/pull/52459 and the issue #1524.
 
 # 6. Optionally, insert a brief introduction for the vote PR, in the markdown format.
 prBody: |
+  Currently, Node.js collaborators are considered inactive if they have not
+  contributed to the project for 12 months. The definition of contribution in
+  the context of Node.js collaborators is any approval or review of a pull
+  request, or any commit to the Node.js repository. The proposed reduction
+  in the duration of inactivity for Node.js collaborators, from 12 months
+  to 9 months.
 
 # 7. Optionally, choose an id that will be used for the branch name as well as
 #    the vote folder name. If not supplied, a UUID will be used.
--- a/votes/initiateNewVote/_EDIT_ME.yml
+++ b/votes/initiateNewVote/_EDIT_ME.yml
@@ -36,7 +36,7 @@ canShuffleCandidates: true
 #    issue it was discussed on to give the full context.
 footerInstructions: |
   The objective of this vote is to decide whether duration to determine
-  inactivity of a from 12 months to 9 months. This vote was discussed
+  inactivity should be reduced from 12 months to 9 months. This vote was discussed
   in pull-request https://github.com/nodejs/node/pull/52459 and the issue #1524.
 
 # 6. Optionally, insert a brief introduction for the vote PR, in the markdown format.
--- a/votes/initiateNewVote/_EDIT_ME.yml
+++ b/votes/initiateNewVote/_EDIT_ME.yml
@@ -35,7 +35,7 @@ canShuffleCandidates: true
 # 5. Insert here a short description of the vote objectives and link to the
 #    issue it was discussed on to give the full context.
 footerInstructions: |
-  The objective of this vote is to decide whether duration to determine
+  The objective of this vote is to decide whether the duration to determine
   inactivity should be reduced from 12 months to 9 months. This vote was discussed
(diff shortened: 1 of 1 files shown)


votes/initiateNewVote/_EDIT_ME.yml
```diff
@@ -35,10 +35,25 @@ canShuffleCandidates: true
...
+  https://github.com/nodejs/node/pull/52459 and the issue #1524.
+
 
 # 6. Optionally, insert a brief introduction for the vote PR, in the markdown format.
 prBody: |
+  Currently, Node.js collaborators are considered inactive if they have not
+  contributed to the project for 12 months. The definition of contribution in
+  the context of Node.js collaborators is any approval or review of a pull
+  request, or any commit to the Node.js repository.
+  The proposed reduction in the duration of inactivity for Node.js collaborators,
```
joyeecheung, line 49: This text is a bit too one-sided (not introducing why some also feel that it is not a good idea). I suggest removing it from the vote intro and you can reply with it to make the case from one side.

votes/initiateNewVote/_EDIT_ME.yml
```diff
@@ -35,10 +35,25 @@ canShuffleCandidates: true
 # 5. Insert here a short description of the vote objectives and link to the
 #    issue it was discussed on to give the full context.
 footerInstructions: |
-  TBD
+  The objective of this vote is to reduce the in-active collaborator duration
```
joyeecheung, line 38:
```suggestion
  The objective of this vote is to decide whether duration to determine inactivity of a collaborator
  should remain as 12 months or should be reduced to 9 months.
```

votes/initiateNewVote/_EDIT_ME.yml
anonrig: Removed it.

votes/initiateNewVote/_EDIT_ME.yml
anonrig: Thanks, applied

votes/initiateNewVote/_EDIT_ME.yml
```diff
@@ -35,10 +35,18 @@ canShuffleCandidates: true
 # 5. Insert here a short description of the vote objectives and link to the
 #    issue it was discussed on to give the full context.
 footerInstructions: |
-  TBD
+  The objective of this vote is to decide whether duration to determine
+  inactivity of a from 12 months to 9 months. This vote was discussed
```
joyeecheung, line 39:
```suggestion
  inactivity should be reduced from 12 months to 9 months. This vote was discussed
```

votes/initiateNewVote/_EDIT_ME.yml
```diff
@@ -35,10 +35,18 @@ canShuffleCandidates: true
...
+  in pull-request https://github.com/nodejs/node/pull/52459 and the issue #1524.
 
 # 6. Optionally, insert a brief introduction for the vote PR, in the markdown format.
 prBody: |
+  Currently, Node.js collaborators are considered inactive if they have not
+  contributed to the project for 12 months. The definition of contribution in
```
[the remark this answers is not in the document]
anonrig: Can you recommend a better wording?

votes/initiateNewVote/_EDIT_ME.yml
```diff
@@ -35,10 +35,18 @@ canShuffleCandidates: true
...
+  the context of Node.js collaborators is any approval or review of a pull
+  request, or any commit to the Node.js repository. The proposed reduction
+  in the duration of inactivity for Node.js collaborators, from 12 months
+  to 9 months.
```
joyeecheung, line 49: I think the current wording is already pretty clear about the specific criteria but if there are still concerns about the definition of contribution not being super duper clear, we can go with

```suggestion
  Currently, Node.js collaborators are considered inactive if in the
  past 12 months, they haven't done any of the following:
  
  - Opening any landed pull request
  - Approving any landed pull request
  - Landing any commit to the Node.js repository
  
  This vote is to decide whether we want to reduce the timespan from 12 months to
  9 months.
```

ruyadorno (approved):

joyeecheung (approved):

votes/initiateNewVote/_EDIT_ME.yml
```diff
@@ -35,10 +35,20 @@ canShuffleCandidates: true
...
 #    issue it was discussed on to give the full context.
 footerInstructions: |
-  TBD
+  The objective of this vote is to decide whether duration to determine
+  inactivity should be reduced from 12 months to 9 months. This vote was discussed
+  in pull-request https://github.com/nodejs/node/pull/52459 and the issue #1524.
 
 # 6. Optionally, insert a brief introduction for the vote PR, in the markdown format.
 prBody: |
+  Currently, Node.js collaborators are considered inactive if in the
```
aduh95, line 44: If you allow me to “Well, technically 🤓”, the proposed change is not how folks are _considered_ (everyone is free to consider someone else as they please, no matter what the TSC decides), it’s about changing the bot behavior. 
```suggestion
  Currently, the bot looks through all commits on `main` that have landed in the last 12 months, and makes two lists:
  
    - a list of all the commit authors.
    - a list of all the collaborators who are listed as reviewers in the commit message (i.e. folks who have approved the PR). 
    
    The bot will open a PR to move a collaborator to emeritus if they don’t appear in either of those lists. 
```

anonrig: looks good to me :-)

anonrig: This broke the text. Can you recommend again?

votes/initiateNewVote/_EDIT_ME.yml
```diff
@@ -35,10 +35,25 @@ canShuffleCandidates: true
...
+  
+    - a list of all the commit authors.
+    - a list of all the collaborators who are listed as reviewers in the commit message (i.e. folks who have approved the PR). 
+    
+    The bot will open a PR to move a collaborator to emeritus if they don’t appear in either of those lists. 
+  past 12 months, they haven't done any of the following:
+  
+  - Authoring any landed commit
+  - Approving any landed commit
+
```
joyeecheung, line 54:
```suggestion
```

This should be replaced by the more concrete list above.

votes/initiateNewVote/_EDIT_ME.yml
```diff
@@ -35,10 +35,25 @@ canShuffleCandidates: true
...
+  in pull-request https://github.com/nodejs/node/pull/52459 and the issue #1524.
 
 # 6. Optionally, insert a brief introduction for the vote PR, in the markdown format.
 prBody: |
+  Currently, the bot looks through all commits on `main` that have landed in the last 12 months, and makes two lists:
+  
+    - a list of all the commit authors.
+    - a list of all the collaborators who are listed as reviewers in the commit message (i.e. folks who have approved the PR). 
+    
+    The bot will open a PR to move a collaborator to emeritus if they don’t appear in either of those lists.
```
joyeecheung, line 49:
```suggestion
  The bot will open a PR to move a collaborator to emeritus if they don’t appear in either of those lists. 
```

(Pretty sure it would look off with the extra indentention).

joyeecheung (approved):

votes/initiateNewVote/_EDIT_ME.yml
joyeecheung: Pro tip: I think even on mobile you can drag on the lines to leave multi-line suggestions

aduh95: Maybe I’m doing it wrong, but it only scrolls for me 🤷‍♂️

aduh95 (approved):
If you’re still pushing for reducing the inactivity period to 6 months, I think you should add it as a candidate (there’s no down side for having more candidates with the vote system we use)

anonrig:
> If you’re still pushing for reducing the inactivity period to 6 months, I think you should add it as a candidate (there’s no down side for having more candidates with the vote system we use)

I think it will make the decision harder. I'll push for 9 for now, and see in the future (maybe in a couple of months)

aduh95:
> I think it will make the decision harder.

I don’t see how. If the voter knows which option they prefer, they can simply order them. If not they can tie the two/three options for which they have no relative preference.

GeoffreyBooth:
I prefer to keep votes simple. Voting to override blocks on a particular PR leaves no ambiguity as to what the outcome of the vote determines: that PR can land.

If we want to consider 6 months, then open a PR for that, let it get blocked, and open a vote to override that block. I'd prefer that process to a ranked choice vote, even though it takes more rounds. Yes it's more inefficient, but we learn information after the first round: how many people support that option and why. That's useful data to have when considering how to vote in later rounds.

votes/initiateNewVote/_EDIT_ME.yml
```diff
@@ -3,7 +3,7 @@
 
 # 1. Select a subject for the vote. This can be a question addressed to the TSC
 #    voting members.
-subject: REPLACEME
+subject: Reduce in-active collaborator duration
```
tniessen, line 6:
```suggestion
subject: Reduce inactive collaborator duration
```

votes/initiateNewVote/_EDIT_ME.yml
```diff
@@ -24,8 +24,8 @@ headerInstructions: |
 #    voters express their preference for each candidates, no matter how many
 #    there are.
 candidates:
-  - TODO
-  - TODO
+  - Status Quo: Keep the current in-active collaborator duration of 12 months
+  - Reduce the in-active collaborator duration to 9 months
```
tniessen, line 28:
```suggestion
  - Status quo: Keep the current inactive collaborator duration of 12 months
  - Reduce the inactive collaborator duration to 9 months
```

votes/initiateNewVote/_EDIT_ME.yml
```diff
@@ -35,10 +35,20 @@ canShuffleCandidates: true
 # 5. Insert here a short description of the vote objectives and link to the
 #    issue it was discussed on to give the full context.
 footerInstructions: |
-  TBD
+  The objective of this vote is to decide whether duration to determine
+  inactivity should be reduced from 12 months to 9 months. This vote was discussed
+  in pull-request https://github.com/nodejs/node/pull/52459 and the issue #1524.
```
tniessen, line 40:
```suggestion
  in pull request https://github.com/nodejs/node/pull/52459 and the issue #1524.
```

votes/initiateNewVote/_EDIT_ME.yml
```diff
@@ -35,10 +35,20 @@ canShuffleCandidates: true
 # 5. Insert here a short description of the vote objectives and link to the
 #    issue it was discussed on to give the full context.
 footerInstructions: |
-  TBD
+  The objective of this vote is to decide whether duration to determine
```
tniessen, line 38:
```suggestion
  The objective of this vote is to decide whether the duration to determine
```

votes/initiateNewVote/_EDIT_ME.yml
```diff
@@ -35,10 +35,20 @@ canShuffleCandidates: true
...
+  in pull-request https://github.com/nodejs/node/pull/52459 and the issue #1524.
 
 # 6. Optionally, insert a brief introduction for the vote PR, in the markdown format.
 prBody: |
+  Currently, the bot looks through all commits on `main` that have landed in the last 12 months, and makes two lists:
+  
+    - a list of all the commit authors.
+    - a list of all the collaborators who are listed as reviewers in the commit message (i.e. folks who have approved the PR). 
+    
+  The bot will open a PR to move a collaborator to emeritus if they don’t appear in either of those lists.
```
tniessen, line 49:
```suggestion
  The bot will open a PR to move a collaborator to emeritus if they don't appear in either of those lists. 
```

MoLow (approved):

aduh95:
> If we want to consider 6 months, then open a PR for that, let it get blocked, and open a vote to override that block

Yagiz already opened that exact PR, and it did get blocked. 

> I prefer to keep votes simple. Voting to override blocks on a particular PR leaves no ambiguity as to what the outcome of the vote determines: that PR can land.

I don’t think adding a third candidate makes the vote less simple, or its outcome more ambiguous. 

In any case, if Yagiz does not want to push for 6 months anymore, that’s certainly fine by me. I’d like the result of the vote to be somewhat final, and not have to discuss the issue again (I mean not until something changes and/or enough[^1] time elapses). 

[^1]: there’s no agreed upon definition of how long is “enough time”.

votes/initiateNewVote/_EDIT_ME.yml
```diff
@@ -35,10 +35,20 @@ canShuffleCandidates: true
...
+  The objective of this vote is to decide whether the duration to determine
+  inactivity should be reduced from 12 months to 9 months. This vote was discussed
+  in pull request https://github.com/nodejs/node/pull/52459 and the issue #1524.
 
 # 6. Optionally, insert a brief introduction for the vote PR, in the markdown format.
 prBody: |
+  Currently, the bot looks through all commits on `main` that have landed in the last 12 months, and makes two lists:
+  
+    - a list of all the commit authors.
+    - a list of all the collaborators who are listed as reviewers in the commit message (i.e. folks who have approved the PR).
```
benjamingr, line 47: I would prefer we add a third option of asking @nodejs/security-wg to review the membership durations and then the TSC to implement their feedback.

I for example would vote against reducing the duration since I feel like it has little security benefit and can alienate people but if I was convinced there would be actual security merit I'd be in favor.

benjamingr (approved):
Approving this since it's always allowed to call for a vote but this seems premature.

benjamingr:
> In any case, if Yagiz does not want to push for 6 months anymore, that’s certainly fine by me. I’d like the result of the vote to be somewhat final, and not have to discuss the issue again (I mean not until something changes and/or enough[1](https://github.com/nodejs/TSC/pull/1536#user-content-fn-1-e62781c40232e6e36c2562f5242a1799) time elapses).

The issue with that is that we don't fully understand the implications yet and calling for a vote now instead of figuring the data out would mean I could more easily vote in some way I'd later regret.

benjamingr:
The motivation is to reduce security risk (right?) in that case I want to know what roles in the org have access to what, what risk it entails and what attack vectors there are.

If a triager is equally as risky or someone can abuse the fact a PR was already OKd once to rerun stuff or if there are other issues (e.g maybe we fix triggers being able to run CI or ensure CI is isolated and then they can still run benchmark ci which isn’t for example).

Basically I would prefer a solution where collaborators can’t do anything that risks us (other than approving PRs obviously) but if we take the limiting inactivity route I at least wanna know it actually helps..

votes/initiateNewVote/_EDIT_ME.yml
```diff
@@ -35,10 +35,20 @@ canShuffleCandidates: true
...
 footerInstructions: |
-  TBD
+  The objective of this vote is to decide whether the duration to determine
+  inactivity should be reduced from 12 months to 9 months. This vote was discussed
+  in pull request https://github.com/nodejs/node/pull/52459 and the issue #1524.
 
 # 6. Optionally, insert a brief introduction for the vote PR, in the markdown format.
 prBody: |
+  Currently, the bot looks through all commits on `main` that have landed in the last 12 months, and makes two lists:
+
```
trivikr, line 45: nit: redundant whitespace
```suggestion

```

benjamingr:
It would be beneficial to add the other options discussed in the TSC meeting here as options

anonrig:
> It would be beneficial to add the other options discussed in the TSC meeting here as options

I'd like to keep the options as it is for now since the original block to my pull request does not involve anything other than 9 months. We can always follow up and vote on the other options if anybody from TSC wants to pursue.

benjamingr:
I mean, we spent time in the TSC meeting discussing several options (reducing it further to 6 months but making it voluntary for example) - merging this as is forced me to vote between two options I don't actually like all while there is ongoing discussion at https://github.com/nodejs/node/pull/52459

PR #1557, opened 2024-05-15
doc: add script used to generate PR for minutes

mhdawson:
This adds the script I used to generate the PR from the google doc after meetings.  You need to be in the meetings directory of a copy of the TSC repo that was checked out. The parameter is the GUID like value in the google meeting doc link.

Adding so that I don't lose it and so other people who Chair TSC meetings can use if they want to as well.


benjamingr (approved):

mcollina (approved):
lgtm

marco-ippolito (approved):

gireeshpunathil (approved):

MoLow (approved):

PR #1574, opened 2024-06-08
Remove `actions/setup-node`

aduh95:
Given we don't need any specific Node.js version in these workflows, we can use the already installed version (https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2204-Readme.md) and get slightly faster CI and fewer things to keep up-to-date.


marco-ippolito (approved):
LGTM assuming you ran it and tested

Issue #1575, opened 2024-06-10
Node.js Technical Steering Committee (TSC) Meeting 2024-06-12


mhdawson:
## Time

**UTC Wed 12-Jun-2024 15:00 (03:00 PM)**:

| Timezone      | Date/Time             |
|---------------|-----------------------|
| US / Pacific  | Wed 12-Jun-2024 08:00 (08:00 AM) |
| US / Mountain | Wed 12-Jun-2024 09:00 (09:00 AM) |
| US / Central  | Wed 12-Jun-2024 10:00 (10:00 AM) |
| US / Eastern  | Wed 12-Jun-2024 11:00 (11:00 AM) |
| EU / Western  | Wed 12-Jun-2024 16:00 (04:00 PM) |
| EU / Central  | Wed 12-Jun-2024 17:00 (05:00 PM) |
| EU / Eastern  | Wed 12-Jun-2024 18:00 (06:00 PM) |
| Moscow        | Wed 12-Jun-2024 18:00 (06:00 PM) |
| Chennai       | Wed 12-Jun-2024 20:30 (08:30 PM) |
| Hangzhou      | Wed 12-Jun-2024 23:00 (11:00 PM) |
| Tokyo         | Thu 13-Jun-2024 00:00 (12:00 AM) |
| Sydney        | Thu 13-Jun-2024 01:00 (01:00 AM) |

Or in your local time:
* https://www.timeanddate.com/worldclock/fixedtime.html?msg=Node.js+Foundation+Technical%20Steering%20Committee%20(TSC)+Meeting+2024-06-12&iso=20240612T1500
* or https://www.wolframalpha.com/input/?i=03PM+UTC%2C+Jun+12%2C+2024+in+local+time

## Links

* Minutes Google Doc: <https://docs.google.com/document/d/1FNdFJfdRXa9dx-FdDpPwlD43uM-HkVIS8ugmNxJv8tI/edit>

## Agenda

Extracted from **tsc-agenda** labelled issues and pull requests from the **nodejs org** prior to the meeting.

### nodejs/node

* module: allow module.register from workers [#53200](https://github.com/nodejs/node/pull/53200)
* src: make `--env-file` return warning when not found [#53177](https://github.com/nodejs/node/pull/53177)
* doc: deprecate NaN or negative number as delay to setTimeout and setInterval [#53005](https://github.com/nodejs/node/pull/53005)
* Ambassador program [#52857](https://github.com/nodejs/node/pull/52857)

[...]

PR #1577, opened 2024-06-14
Propose vote to unblock `--env-file` PRs

aduh95:
Refs: https://github.com/nodejs/TSC/issues/1575#issuecomment-2166976221

### DIFF
--- a/votes/initiateNewVote/_EDIT_ME.yml
+++ b/votes/initiateNewVote/_EDIT_ME.yml
@@ -42,11 +42,11 @@ footerInstructions: |
   Based on the result of this vote, we can unblock either https://github.com/nodejs/node/pull/53177,
   https://github.com/nodejs/node/pull/53060, or neither if the status quo prevails. The name of the
   proposed flags are non-binding, and are provided only as examples. If a proposal that includes a
-  new flag, the name of that flag can be bike-shedded per the PR review process.
+  new flag is accepted, the name of that flag can be bike-shed per the PR review process.
 
 # 6. Optionally, insert a brief introduction for the vote PR, in the markdown format.
 prBody: |
-  The vote should stay open for 7 days, or until all TSC voting members have case a ballot.
+  The vote should stay open for 7 days, or until all TSC voting members have cast a ballot.
 
 # 7. Optionally, choose an id that will be used for the branch name as well as
 #    the vote folder name. If not supplied, a UUID will be used.
--- a/votes/initiateNewVote/_EDIT_ME.yml
+++ b/votes/initiateNewVote/_EDIT_ME.yml
@@ -28,7 +28,8 @@ candidates:
   - Ignore when passed to `--env-file` flag, provide no other flag.
   - Warn when passed to `--env-file` flag, provide no other flag.
   - Throw when passed to `--env-file` flag, warn when passed to a new `--env-file-if-exists`.
-  - Throw when passed to `--env-file` flag, do nothing when passed to a new `--env-file-if-exists`.
+  - Throw when passed to `--env-file` flag, ignore when passed to a new `--env-file-if-exists`.
+  - Ignore when passed to `--env-file` flag, throw when passed to a new `--env-file-must-exist`.
   - Throw when passed to `--env-file` flag, unless new `--env-file-when-missing=warn` is also passed.
   - Warn when passed to `--env-file` flag, unless new `--env-file-when-missing=throw` is also passed.
 
@@ -44,6 +45,7 @@ footerInstructions: |
   https://github.com/nodejs/node/pull/53060, or neither if the status quo prevails. The name of the
   proposed flags are non-binding, and are provided only as examples. If a proposal that includes a
   new flag is accepted, the name of that flag can be bike-shed per the PR review process.
+  "Ignore" means "do not throw and do not warn" in this context.
 
 # 6. Optionally, insert a brief introduction for the vote PR, in the markdown format.
 prBody: |
--- a/votes/initiateNewVote/_EDIT_ME.yml
+++ b/votes/initiateNewVote/_EDIT_ME.yml
@@ -3,7 +3,7 @@
 
 # 1. Select a subject for the vote. This can be a question addressed to the TSC
 #    voting members.
-subject: REPLACEME
+subject: What UX do we want `node` to handle non-existent `.env` files passed to its CLI?
 
 # 2. You can leave the header instructions as is, or modify them if you see fit.
 headerInstructions: |
@@ -24,8 +24,12 @@ headerInstructions: |
 #    voters express their preference for each candidates, no matter how many
 #    there are.
 candidates:
-  - TODO
-  - TODO
+  - Throw when passed to `--env-file` flag, provide no other flag (status quo).
+  - Warn when passed to `--env-file` flag, provide no other flag.
+  - Throw when passed to `--env-file` flag, warn when passed to a new `--env-file-if-exists`.
+  - Throw when passed to `--env-file` flag, do nothing when passed to a new `--env-file-if-exists`.
+  - Throw when passed to `--env-file` flag, unless new `--env-file-when-missing=warn` is also passed.
+  - Warn when passed to `--env-file` flag, unless new `--env-file-when-missing=throw` is also passed.
 
 # 4. Pass the following to false if it's important to keep the candidates in the
 #    order you define above. Presenting candidates in a fixed order tends to
@@ -35,11 +39,15 @@ canShuffleCandidates: true
 # 5. Insert here a short description of the vote objectives and link to the
 #    issue it was discussed on to give the full context.
 footerInstructions: |
-  TBD
+  Based on the result of this vote, we can unblock either https://github.com/nodejs/node/pull/53177,
+  https://github.com/nodejs/node/pull/53060, or neither if the status quo prevails. The name of the
+  proposed flags are non-binding, and are provided only as examples. If a proposal that includes a
+  new flag, the name of that flag can be bike-shedded per the PR review process.
 
 # 6. Optionally, insert a brief introduction for the vote PR, in the markdown format.
 prBody: |
+  The vote should stay open for 7 days, or until all TSC voting members have case a ballot.
 
 # 7. Optionally, choose an id that will be used for the branch name as well as
 #    the vote folder name. If not supplied, a UUID will be used.
-path-friendly-id: null
+path-friendly-id: how-to-handle-missing-env-files
--- a/votes/initiateNewVote/_EDIT_ME.yml
+++ b/votes/initiateNewVote/_EDIT_ME.yml
@@ -3,7 +3,7 @@
 
 # 1. Select a subject for the vote. This can be a question addressed to the TSC
 #    voting members.
-subject: What UX do we want `node` to handle non-existent `.env` files passed to its CLI?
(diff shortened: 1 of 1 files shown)


mcollina (approved):
lgtm

marco-ippolito (approved):

gireeshpunathil (approved):

votes/initiateNewVote/_EDIT_ME.yml
```diff
@@ -24,8 +24,12 @@ headerInstructions: |
...
 #    there are.
 candidates:
-  - TODO
-  - TODO
+  - Throw when passed to `--env-file` flag, provide no other flag (status quo).
+  - Warn when passed to `--env-file` flag, provide no other flag.
+  - Throw when passed to `--env-file` flag, warn when passed to a new `--env-file-if-exists`.
+  - Throw when passed to `--env-file` flag, do nothing when passed to a new `--env-file-if-exists`.
+  - Throw when passed to `--env-file` flag, unless new `--env-file-when-missing=warn` is also passed.
+  - Warn when passed to `--env-file` flag, unless new `--env-file-when-missing=throw` is also passed.
```
GeoffreyBooth, line 32: I don't think anyone has been advocating for these. These options would mean you can't have a mix of required and optional files to load.

aduh95: > I don't think anyone has been advocating for these.

I have


> These options would mean you can't have a mix of required and optional files to load.

That's correct

votes/initiateNewVote/_EDIT_ME.yml
```diff
@@ -3,7 +3,7 @@
 
 # 1. Select a subject for the vote. This can be a question addressed to the TSC
 #    voting members.
-subject: REPLACEME
+subject: What UX do we want `node` to handle non-existent `.env` files passed to its CLI?
```
tniessen, line 6: I don't have a suggestion but something in this question seems off, linguistically.

votes/initiateNewVote/_EDIT_ME.yml
```diff
@@ -35,11 +39,15 @@ canShuffleCandidates: true
 # 5. Insert here a short description of the vote objectives and link to the
 #    issue it was discussed on to give the full context.
 footerInstructions: |
-  TBD
+  Based on the result of this vote, we can unblock either https://github.com/nodejs/node/pull/53177,
+  https://github.com/nodejs/node/pull/53060, or neither if the status quo prevails. The name of the
+  proposed flags are non-binding, and are provided only as examples. If a proposal that includes a
+  new flag, the name of that flag can be bike-shedded per the PR review process.
```
tniessen, line 45:
```suggestion
  proposed flags are non-binding, and are provided only as examples. If a proposal that includes a
  new flag is accepted, the name of that flag can be bike-shed per the PR review process.
```

votes/initiateNewVote/_EDIT_ME.yml
```diff
@@ -35,11 +39,15 @@ canShuffleCandidates: true
...
 
 # 6. Optionally, insert a brief introduction for the vote PR, in the markdown format.
 prBody: |
+  The vote should stay open for 7 days, or until all TSC voting members have case a ballot.
```
tniessen, line 49:
```suggestion
  The vote should stay open for 7 days, or until all TSC voting members have cast a ballot.
```

tniessen (approved):

MoLow (approved):

votes/initiateNewVote/_EDIT_ME.yml
```diff
@@ -24,8 +24,12 @@ headerInstructions: |
 #    voters express their preference for each candidates, no matter how many
 #    there are.
 candidates:
-  - TODO
-  - TODO
+  - Throw when passed to `--env-file` flag, provide no other flag (status quo).
```
benjamingr, line 27: Can we add another option

```suggestion
  - Do nothing (this is the same behavior as `dotenv` and every other library and programming language supporting this).
```

(Assuming there isn't another library or language with different behavior I'm just not aware of)

benjamingr: (I'm also happy with different phrasing)

anonrig: I don't think this is correct. `dotenv` does not throw error if file is not found, but we throw an error right now.

anonrig: Can you rephrase "do nothing" to "don't warn and don't throw an error"?

aduh95:
```suggestion
  - Throw when passed to `--env-file` flag, provide no other flag (status quo).
  - Ignore when passed to `--env-file` flag, provide no other flag.
```

anonrig:
> I really dislike the project landing a feature it's adopting because it reached consensus in userland and then changing its behavior by committee even though all userland libraries (at least popular ones I found) have the same behavior (which is different from what we're proposing).

Note: I agree with @benjamingr 100%.

votes/initiateNewVote/_EDIT_ME.yml
BoscoDomingo: Yes, I agree. To me it should be "How should `node` handle non-existent `.env` files passed via CLI"

votes/initiateNewVote/_EDIT_ME.yml
anonrig: I think we should make it clear that: `(this is the same behavior as `dotenv` and other libraries and platforms supporting .env)`

votes/initiateNewVote/_EDIT_ME.yml
aduh95:
```suggestion
subject: How should `node` handle non-existent `.env` files passed via CLI flags?
```

ruyadorno (approved):

votes/initiateNewVote/_EDIT_ME.yml
benjamingr: I am fine with any phrasing that makes it explicit that this option is the one that isn't "we picked an API that reached maturity through userland and we're sticking with the design our community adopted rather than deciding to change its design by committee".

I honestly don't have a strong opinion about whether we should throw or warn or ignore by default but I trust our community and ecosystem to adopt the option with better developer experience. The fact the ~10 implementations I checked* reached the same conclusion and it's different from all our options is troubling to me.


* (and I may be wrong and there are others who reached a different conclusion)

BoscoDomingo:
> > I really dislike the project landing a feature it's adopting because it reached consensus in userland and then changing its behavior by committee even though all userland libraries (at least popular ones I found) have the same behavior (which is different from what we're proposing).
> 
> Note: I agree with @benjamingr 100%.

I agree as well, we just each approached the problem from a different angle. In the end we both want missing .env files to not throw, I just made it a decision to be taken by the user, whereas you prefer sticking to other tools' existing behaviour, and that's 100% fine too 👍🏼 

Also, the proposal LGTM, thanks @aduh95

BoscoDomingo (approved):

votes/initiateNewVote/_EDIT_ME.yml
benjamingr: This is and has been the behavior for a while though you can add a config option (verbose:true or debug:true) in order to change it to warn. At least that is my understanding

votes/initiateNewVote/_EDIT_ME.yml
```diff
@@ -24,8 +24,13 @@ headerInstructions: |
...
+  - Ignore when passed to `--env-file` flag, provide no other flag.
+  - Warn when passed to `--env-file` flag, provide no other flag.
+  - Throw when passed to `--env-file` flag, warn when passed to a new `--env-file-if-exists`.
+  - Throw when passed to `--env-file` flag, do nothing when passed to a new `--env-file-if-exists`.
```
aduh95, line 31: For the sake of completeness 
```suggestion
  - Throw when passed to `--env-file` flag, ignore when passed to a new `--env-file-if-exists`.
  - Ignore when passed to `--env-file` flag, throw when passed to a new `--env-file-must-exist`.
```

benjamingr:
Basically my stance isn’t warn, ignore or error - it’s “do whatever userland does”.

votes/initiateNewVote/_EDIT_ME.yml
```diff
@@ -35,11 +40,15 @@ canShuffleCandidates: true
 # 5. Insert here a short description of the vote objectives and link to the
 #    issue it was discussed on to give the full context.
 footerInstructions: |
-  TBD
+  Based on the result of this vote, we can unblock either https://github.com/nodejs/node/pull/53177,
+  https://github.com/nodejs/node/pull/53060, or neither if the status quo prevails. The name of the
+  proposed flags are non-binding, and are provided only as examples. If a proposal that includes a
+  new flag is accepted, the name of that flag can be bike-shed per the PR review process.
```
aduh95, line 46:
```suggestion
  new flag is accepted, the name of that flag can be bike-shed per the PR review process.
  "Ignore" means "do not throw and do not warn" in this context.
```

mhdawson (approved):
LGTM

jasnell (approved):

votes/initiateNewVote/_EDIT_ME.yml
GeoffreyBooth: Sorry I missed that. I feel like a bunch of people would be upset if this is what we land, because the use case they want to support is “I have some env files that every environment needs, plus another env file that only localhost needs”. That was what prompted `--env-file-optional` if I remember correctly.

aduh95: > I feel like a bunch of people would be upset if this is what we land

Yes probably, and that applies to all the candidates BTW – that's why we're having a vote, because we couldn't find a solution that would please everybody.

GeoffreyBooth:
> Basically my stance isn’t warn, ignore or error - it’s “do whatever userland does”.

According to what you’ve written, though, `dotenv` has a default behavior that can be overridden by its config file. We don’t have config file support, though, so at the moment we can’t really mimic `dotenv`‘s behavior, right?

I guess that could be another option, though: add [support for a general Node config file](https://github.com/nodejs/node/pull/53177#issuecomment-2146198245), and then `--env-file` has the warn behavior by default and it can be set to either throw or be silent via the config file. That would exactly match `dotenv`, I think?

benjamingr:
@GeoffreyBooth the equivalent would be "do not warn on missing .env file, add an option to process.loadEnvFile to warn/throw on a missing env file

GeoffreyBooth:
> the equivalent would be “do not warn on missing .env file, add an option to process.loadEnvFile to warn/throw on a missing env file

That’s not equivalent to `--env-file-optional` or `--env-file-required` because by the time you can run `process.loadEnvFile` or `util.parseEnv`, Node is already running and therefore any `NODE_OPTIONS` within the `.env` file has no effect. Part of the existing use case for `--env-file` is as a way to define Node configuration via `NODE_OPTIONS`.

benjamingr:
> That’s not equivalent to --env-file-optional or --env-file-required because by the time you can run process.loadEnvFile or util.parseEnv, Node is already running and therefore any NODE_OPTIONS within the .env file has no effect.

Correct, I was only explaining what userland does and was suggesting we keep the same existing behavior that reached consensus in our ecosystem over many years rather than change it based on how a technical committee believes things should work.

PR #1621, opened 2024-09-18
Start a vote on require(esm) markers

joyeecheung:
This is a follow-up to the naming bikeshed in https://github.com/nodejs/loaders/issues/221 which is for the implementation in https://github.com/nodejs/node/pull/54563 A summary can be found in https://github.com/nodejs/node/pull/54563#pullrequestreview-2288424604

### DIFF
--- a/votes/initiateNewVote/_EDIT_ME.yml
+++ b/votes/initiateNewVote/_EDIT_ME.yml
@@ -90,7 +90,7 @@ prBody: |
     module.exports = 'bar';
   ```
 
-  Users importing with `console.log(await import('bar'))` the user would a namespace object either of the form `ModuleNamespace { __cjsUnwrapDefault: true, default: 'bar' }` or `ModuleNamespace { default: 'bar', 'module.exports': 'bar' }` depending on which option is chosen for this first PR.
+  Users importing with `console.log(await import('bar'))` would obtain a namespace object either of the form `ModuleNamespace { __cjsUnwrapDefault: true, default: 'bar' }` or `ModuleNamespace { default: 'bar', 'module.exports': 'bar' }` depending on which option is chosen for this first PR.
 
   The vote should stay open for 7 days, or until all TSC voting members have cast a ballot.
 
--- a/votes/initiateNewVote/_EDIT_ME.yml
+++ b/votes/initiateNewVote/_EDIT_ME.yml
@@ -65,7 +65,9 @@ prBody: |
   const foo = require('foo').default;
   ```
 
-  As you can see, for CommonJS consumers this will not be exactly equivalent. So the idea is that Node.js can take a marker from package authors to customize what `require(esm)` gets in this case. To keep `const foo = require('foo')` working, the two proposed markers are:
+  As you can see, for CommonJS consumers this will not be exactly equivalent. Since in ESM the default exports is a property of the name space, they need to access `.default` from the result returned by
+  `require(esm)`. So the idea is that Node.js can take a marker from package authors to directly customize what `require(esm)` returns in this case.
+  To keep `const foo = require('foo')` working, the two proposed markers are:
 
   ```js
   export default function foo () { };
--- a/votes/initiateNewVote/_EDIT_ME.yml
+++ b/votes/initiateNewVote/_EDIT_ME.yml
@@ -37,6 +37,7 @@ canShuffleCandidates: true
 footerInstructions: |
   This is a follow-up to the naming bikeshed in https://github.com/nodejs/loaders/issues/221
   which is for the implementation in https://github.com/nodejs/node/pull/54563
+  and the follow-on issue in https://github.com/nodejs/node/pull/53848
   A short summary can be found in https://github.com/nodejs/node/pull/54563#pullrequestreview-2288424604
 
 # 6. Optionally, insert a brief introduction for the vote PR, in the markdown format.
--- a/votes/initiateNewVote/_EDIT_ME.yml
+++ b/votes/initiateNewVote/_EDIT_ME.yml
@@ -82,9 +82,15 @@ prBody: |
   export { foo as "module.exports" };
   ```
 
-  Note that to follow up with this vote, there is [a proposal](https://github.com/nodejs/node/pull/53848) to reuse this marker for detecting whether the imported module is CJS.
-  There is no consensus on whether that is the right way to solve the problem yet ([see](https://github.com/nodejs/node/pull/53848#issuecomment-2244500584)), so this vote is just about
-  the marker naming bikeshed.
+  Further to this vote, there is [a follow-on proposal](https://github.com/nodejs/node/pull/53848) to reuse this marker on namespaces representing CJS modules imported into ESM, where attaching this marker would effectively define a scheme for converting CJS into ESM that would retain full transitive interop semantics allowing arbitrary CJS to be converted into ESM as a non-breaking change.
+
+  Under this follow-on marking, when importing some CJS module `bar` into ESM:
+
+  ```js
+    module.exports = 'bar';
+  ```
+
+  Users importing with `console.log(await import('bar'))` the user would a namespace object either of the form `ModuleNamespace { __cjsUnwrapDefault: true, default: 'bar' }` or `ModuleNamespace { default: 'bar', 'module.exports': 'bar' }` depending on which option is chosen for this first PR.
 
   The vote should stay open for 7 days, or until all TSC voting members have cast a ballot.
 
--- a/votes/initiateNewVote/_EDIT_ME.yml
+++ b/votes/initiateNewVote/_EDIT_ME.yml
@@ -62,7 +62,7 @@ prBody: |
   // ESM users get...
   import foo from 'foo';
   // CommmonJS users get..
-  const foo = require('foo').default;
+  const foo = require('foo').default;   // <--- This is now different!
   ```
 
   As you can see, for CommonJS consumers this will not be exactly equivalent. Since in ESM the default exports is a property of the name space, they need to access `.default` from the result returned by
--- a/votes/initiateNewVote/_EDIT_ME.yml
+++ b/votes/initiateNewVote/_EDIT_ME.yml
@@ -82,6 +82,10 @@ prBody: |
   export { foo as "module.exports" };
   ```
 
+  Note that to follow up with this vote, there is [a proposal](https://github.com/nodejs/node/pull/53848) to reuse this marker for detecting whether the imported module is CJS.
+  There is no consensus on whether that is the right way to solve the problem yet ([see](https://github.com/nodejs/node/pull/53848#issuecomment-2244500584)), so this vote is just about
+  the marker naming bikeshed.
+
   The vote should stay open for 7 days, or until all TSC voting members have cast a ballot.
 
 # 7. Optionally, choose an id that will be used for the branch name as well as
--- a/votes/initiateNewVote/_EDIT_ME.yml
+++ b/votes/initiateNewVote/_EDIT_ME.yml
@@ -3,7 +3,7 @@
 
 # 1. Select a subject for the vote. This can be a question addressed to the TSC
 #    voting members.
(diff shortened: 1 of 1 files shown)


joyeecheung:
cc @nodejs/tsc @nodejs/loaders @guybedford

votes/initiateNewVote/_EDIT_ME.yml
```diff
@@ -35,10 +35,53 @@ canShuffleCandidates: true
 # 5. Insert here a short description of the vote objectives and link to the
 #    issue it was discussed on to give the full context.
 footerInstructions: |
-  TBD
+  This is a follow-up to the naming bikeshed in https://github.com/nodejs/loaders/issues/221
+  which is for the implementation in https://github.com/nodejs/node/pull/54563
+  A short summary can be found in https://github.com/nodejs/node/pull/54563#pullrequestreview-2288424604
```
guybedford, line 40:
```suggestion
  This is a follow-up to the naming bikeshed in https://github.com/nodejs/loaders/issues/221
  which is for the implementation in https://github.com/nodejs/node/pull/54563
  and the follow-on issue in https://github.com/nodejs/node/pull/53848
  A short summary can be found in https://github.com/nodejs/node/pull/54563#pullrequestreview-2288424604
```

votes/initiateNewVote/_EDIT_ME.yml
```diff
@@ -35,10 +35,53 @@ canShuffleCandidates: true
...
+  export const __cjsUnwrapDefault = true;
+  ```
+
+  or
+
+  ```js
+  export default function foo () { };
+  export { foo as "module.exports" };
+  ```
+
```
guybedford, line 83: I'd like to include a description of the follow-on work here;

```suggestion
  As a follow-up feature, on the other side, when importing some CJS module `bar` into ESM:
  
  bar.js
  ```js
    module.exports = 'bar';
  ```js
  
  Users importing with `console.log(await import('bar'))` the user will get a namespace object of the form `ModuleNamespace { default: 'bar' }`.
  
  To support stable transitive interop patterns that can work both ways, the same flag can be exported on the namespace object when CJS modules are represented in ESM - either `ModuleNamespace { __cjsUnwrapDefault: true, default: 'bar' }` or `ModuleNamespace { default: 'bar', 'module.exports': 'bar' }` depending on which option is chosen for this first PR.
```

Note the closing code sections need to have the js suffix removed.

guybedford (changes requested):
Thanks @joyeecheung for putting this together. Just added some extra notes on the add-on PR here, as it is also part of the implications here.

votes/initiateNewVote/_EDIT_ME.yml
joyeecheung: I don't think we have consensus about using the namespace for this, and that might also come down to a vote, but I can leave a link about the potential followup.

votes/initiateNewVote/_EDIT_ME.yml
```diff
@@ -3,7 +3,7 @@
 
 # 1. Select a subject for the vote. This can be a question addressed to the TSC
 #    voting members.
-subject: REPLACEME
+subject: Choosing a marker to customize default exports returned by require(esm)
```
aduh95, line 6:
```suggestion
subject: Choosing a marker to customize default exports returned by `require(esm)`
```

votes/initiateNewVote/_EDIT_ME.yml
```diff
@@ -24,8 +24,8 @@ headerInstructions: |
 #    voters express their preference for each candidates, no matter how many
 #    there are.
 candidates:
-  - TODO
-  - TODO
+  - export { foo as "module.exports" };
+  - export const __cjsUnwrapDefault = true;
```
aduh95, line 28:
```suggestion
  - '`export { foo as "module.exports" };`'
  - '`export const __cjsUnwrapDefault = true;`'
```

votes/initiateNewVote/_EDIT_ME.yml
guybedford: @joyeecheung when we first discussed this over two months ago now you explicitly told me you were okay with this, and now you are saying you might not only hinder this first phase compromise but potentially also block the second part of my PR. We need to do better as collaborators to not waste eachother's time.

votes/initiateNewVote/_EDIT_ME.yml
```diff
@@ -35,10 +35,58 @@ canShuffleCandidates: true
...
+  or
+
+  ```js
+  export default function foo () { };
+  export { foo as "module.exports" };
+  ```
+
+  Note that to follow up with this vote, there is [a proposal](https://github.com/nodejs/node/pull/53848) to reuse this marker for detecting whether the imported module is CJS.
+  There is no consensus on whether that is the right way to solve the problem yet ([see](https://github.com/nodejs/node/pull/53848#issuecomment-2244500584)), so this vote is just about
+  the marker naming bikeshed.
```
guybedford, line 87: This is not resolved, and my request for including the explicit shape of the namespace remains as per my comment.

guybedford (changes requested):

votes/initiateNewVote/_EDIT_ME.yml
joyeecheung: If I am looking at the timelines correctly, it went like this:

1. I was okay with it at first, because I thought it was not too much too add
2. I then realized that this would be semver-major since it would break the namespace, which I had not really realized before, so the "not too much to add" is no longer valid. And then, I realized we could potentially solve the use case without breaking the namespace with e.g. https://github.com/nodejs/node/pull/53848#issuecomment-2244500584

I think in this project, we are allowed to change our minds when we have new discoveries about the implications of a proposal. At the end of the day, when there are disagreements, we go to a vote. I am fine with breaking the namespace if the majority of the group supports it, but I don't feel comfortable to say "I am fine with it because I had said so when I did not realize there were more implications & alternatives, and I would just pretend that I didn't have any of the realizations to speed this up, and let the decision be made by just these two people".

Sorry if you feel that your time has been wasted. I believe we are just trying to make module loading better in Node.js, even though we have different priorities.

votes/initiateNewVote/_EDIT_ME.yml
joyeecheung: Isn't the referenced link https://github.com/nodejs/node/pull/53848 already including that proposed namespace?

ShogunPanda (approved):

votes/initiateNewVote/_EDIT_ME.yml
guybedford: The issue is that you've put me in a situation where I need a vote on PR to land my compromise to then land my other compromise of a major change (with the delays now causing us to miss the 23 release milestone).

There is a very real impact on progress that comes with this behaviour, and while you have every right to object, I have every right to feel annoyed!

votes/initiateNewVote/_EDIT_ME.yml
guybedford: The wording here is really dismissive to the use case I've been working on for two months @joyeecheung which is the reason I'm working on all this, which I really don't appreciate.

Transitive interop patterns are important and a key feature of this flag for the future of transpilation of CommonJS into ESM. The vote description absolutely needs to cover this aspect and the use case I'm working on here.

Yes we have different use cases, and I've compromised my use cases to solve yours, but you cannot go around dismissing my use case discussions as abstract hypothetical future scenarios, it's plainly rude as written here.

To correct this line:

```suggestion
  Further to this vote, there is [a follow-on proposal](https://github.com/nodejs/node/pull/53848) to reuse this marker on namespaces representing CJS modules imported into ESM, where attaching this marker would effectively define a scheme for converting CJS into ESM that would retain full transitive interop semantics allowing arbitrary CJS to be converted into ESM as a non-breaking change.
```

My suggestion remains:

```js
Under this follow-on marking, when importing some CJS module `bar` into ESM:
  
  bar.js
  ```js
    module.exports = 'bar';
  ```js
  
  Users importing with `console.log(await import('bar'))` the user would a namespace object either of the form `ModuleNamespace { __cjsUnwrapDefault: true, default: 'bar' }` or `ModuleNamespace { default: 'bar', 'module.exports': 'bar' }` depending on which option is chosen for this first PR.
```

joyeecheung: I am not sure what part about the wording is rude, or dismissive. I am just trying to keep the description on-point and leave the additional context to a link.

>  Note that to follow up with this vote, there is [a proposal](https://github.com/nodejs/node/pull/53848) to reuse this marker for detecting whether the imported module is CJS.

I think this is just explaining the situation as it is, without derailing the vote with another topic.

>  There is no consensus on whether that is the right way to solve the problem yet ([see](https://github.com/nodejs/node/pull/53848#issuecomment-2244500584)), so this vote is just about
  the marker naming bikeshed.

I think this is just describing the fact? We don't have consensus that the reuse is the right way for `import(cjs)` detection, and this vote is not about that, either?

But anyway, if you insist on including the additional information and making the text longer by about 800+ words, I can include them. Personally I think the longer the text is, the harder it is for voters to digest, or at least when I am asked to vote, I would hope that the summary can just cut to the chase and keep the additional context in links. But there is no need to let the vote get stuck here.

votes/initiateNewVote/_EDIT_ME.yml
joyeecheung: Technically, I didn't formally block the PRs (with request changes), which was also what I said 2 months ago - I would not block them but I was also skeptical about them - so when you have enough approvals after the final commit, you could land them despite my concerns. If I didn't want progress to be made, I would not suggest to vote in the first place - the TSC vote is specifically used to prevent a standstill and come to a decision.

I understand that maybe in TC39 or the former modules group at some point, there is a "one person disagrees and it will never proceed" situation, and that might lead you to believe that Node.js had the same model and you absolutely needed my approval to land the PRs. But no, in Node.js, you could either use a vote to accelerate the decision making, or just land them despite having non-blocking concerns at your own discretion.

votes/initiateNewVote/_EDIT_ME.yml
```diff
@@ -35,10 +35,64 @@ canShuffleCandidates: true
...
+
+  Further to this vote, there is [a follow-on proposal](https://github.com/nodejs/node/pull/53848) to reuse this marker on namespaces representing CJS modules imported into ESM, where attaching this marker would effectively define a scheme for converting CJS into ESM that would retain full transitive interop semantics allowing arbitrary CJS to be converted into ESM as a non-breaking change.
+
+  Under this follow-on marking, when importing some CJS module `bar` into ESM:
+
+  ```js
+    module.exports = 'bar';
+  ```
+
+  Users importing with `console.log(await import('bar'))` the user would a namespace object either of the form `ModuleNamespace { __cjsUnwrapDefault: true, default: 'bar' }` or `ModuleNamespace { default: 'bar', 'module.exports': 'bar' }` depending on which option is chosen for this first PR.
```
guybedford, line 93: Minor typo from my suggestion before:

```suggestion
  Users importing with `console.log(await import('bar'))` would obtain a namespace object either of the form `ModuleNamespace { __cjsUnwrapDefault: true, default: 'bar' }` or `ModuleNamespace { default: 'bar', 'module.exports': 'bar' }` depending on which option is chosen for this first PR.
```

guybedford (approved):
Thanks @joyeecheung for taking on board the feedback, happy to go ahead, and as I say I completely commit to following whatever is voted on here.

mhdawson (approved):
LGTM

joyeecheung:
I am not entirely sure how long this should wait, https://github.com/nodejs/TSC/tree/main/votes/initiateNewVote#initiate-a-new-tsc-vote just says "Once the PR has approvals", and it's been open for about 2 days with a few approvals, so I guess I'll just merge it when it passes the 48-hour mark.

marco-ippolito (approved):

MoLow (approved):

PR #1641, opened 2024-10-26
doc: clarify regular member expectation

marco-ippolito:
Refs: https://github.com/nodejs/node/pull/55524
This PR wants to set the right expectation on regular membership:
- A regular member can be asked to be moved to emeritus if inactive for extended period of time
- A regular member is a temporary status
- Emeritus member can be moved back to voting member if there is consensus otherwise will need to be re nominated

This makes it easier to be moved to emeritus, without having to stay as a regular member indefinitely.

Otherwise regular membership its a weird status where someone does not fullfil the role of TSC but does not lose the ability of regaining it even after prolonged period of inactivity.

@nodejs/tsc

### DIFF
--- a/TSC-Charter.md
+++ b/TSC-Charter.md
@@ -55,6 +55,13 @@ TSC voting members are expected to regularly participate in TSC activities.
 
 A TSC voting member is automatically converted to a TSC regular member if
 they do not participate in three consecutive TSC votes.
+Regular membership is expected to be a temporary status, with the
+expectation that the regular member will return to be active and be
+converted to voting members after a period of time.
+A TSC regular member can be asked to become emeritus in the event that they
+are not able to participate in TSC activities for an extended period of time.
+An emeritus member can ask to be reactivated as a regular member by a standard TSC
+motion, otherwise, they will need to be re-elected as a voting member.
 
 ## Section 4. Responsibilities of the TSC
 


benjamingr (approved):
This seems reasonable to me.

TSC-Charter.md
```diff
@@ -55,6 +55,13 @@ TSC voting members are expected to regularly participate in TSC activities.
 
 A TSC voting member is automatically converted to a TSC regular member if
 they do not participate in three consecutive TSC votes.
+Regular membership is expected to be a temporary status, with the
```
anonrig, line 58: Can we give a specific duration?

marco-ippolito: Maybe match the collaborator inactivity duration (I think was 9 months)

aduh95 (commented):
IMO that's not necessary, there's already a requirement for a TSC member to be a collaborator, so if they are inactive for 9 months, there are moved to emeritus already. tbh I don't see a problem with having folks as regular members indefinitely, in fact it was the point when this status was introduced.

marco-ippolito:
What is the point of someone in a leadership position, who is not actively involved in leadership and has not plan to go back to that position? If it worked like collaborators then there should not even exist the "regular member" status.
If its not temporary it should be definitive. I'm ok with not setting a hard limit.

aduh95:
Well the regular status is relatively recent, and the change you’re suggesting is kinda regressing to the previous situation with more steps. If that’s what we want to do, I’d rather have us remove that regular status than making it temporary. 

You ask what is the point: it’s the least resistance path. AKAIK there’s no downsides for having regular TSC members indefinitely, so there’s no reason add a time limit. I’m open for being convinced otherwise, but until then I’m -1.

ShogunPanda (approved):
LGTM!

anonrig (commented):
Can we talk with regular members since regular members include long time valuable contributors. I feel like it would be rude to them if we made this decision without at least communicating their intentions, or their interest in rejoining the TSC members. On top of async talks between all parties, we can talk about this on the collaborator summit next week. 

PS: I think we should try to earn regular members, and I feel like this is a step backwards.

Trott:
> What is the point of someone in a leadership position, who is not actively involved in leadership and has not plan to go back to that position?

An excellent question. The answer is: This is not actually a leadershp position. It is a ceremonial role.

Here's what happened: We had inactive TSC members game the requirements for years to stay on as a TSC member. They wanted the title. I don't know if it was for general status, just their own sense of self, a sense that being a Node.js TSC member would enhance their career prospects, or what. But some people really didn't want to let go and the TSC has always avoided conflict. So no one was ever voted off the TSC except for exactly one time when I believe they had been completely absent and non-responsive for 4 years.

There's a lot I'm leaving out about the impossibility of enforcing the original activity rules, and attempts to iterate them until they were effective. But the point is this:

For whatever reason, there are people who want to be a TSC member but don't or can't do much or any of the TSC work. We spent years--literally years!--alternately ignoring this problem and then trying different things that invariably failed. The only thing that worked was saying "OK, fine, we'll call you a 'regular member', which we know means 'inactive' but that's OK. As long as you stay a Node.js collaborator, you can keep your TSC title. We know that 'voting member' means The Actual TSC.™"

Obviously, theis arrangement has its own drawbacks. It's confusing, for one thing. And perhaps it has outlived its usefulness. But honestly, if it's just bothering someone but not causing any actual measurable and specific problems, I'd recommend just leaving it alone. It's not without problems, but it is probably the best setup we've had. Moving someone to a "regular" member is not nearly the heavy lift of moving them to "emeritus". That may be hard to believe, but it's true, or at least was true.

All that said, I'm fine with whatever you want to do. I'm just providing the backstory here because I'm the person that created this messy setup and I know why it is the way it is.

TSC-Charter.md
```diff
@@ -55,6 +55,13 @@ TSC voting members are expected to regularly participate in TSC activities.
...
+expectation that the regular member will return to be active and be
+converted to voting members after a period of time.
+A TSC regular member can be asked to become emeritus in the event that they
+are not able to participate in TSC activities for an extended period of time.
+An emeritus member can ask to be reactivated as a regular member by a standard TSC
+motion, otherwise, they will need to be re-elected as a voting member.
```
Trott, line 64: I'm confused by this. Did something change? Bringing on a new TSC voting member was always a standard motion/vote within the TSC. This makes it sound like there's a different nominating process. Maybe I missed a change in the last year or so.

anonrig: Q: What is the difference between standard motion and a vote?

aduh95: The same difference between lazy consensus (the motion is accepted as long as no one objects) and a formal vote (where all voting members are expected to participate or explicitly abstain)

marco-ippolito: Maybe its `re-elected` is not the right word, but its always a standard motion (lazy consensus)

mhdawson:
@Trott thanks for the backstory.

I see the TLDR; as it is hard to disengage after being a TSC member and the regular member status makes that transition a bit easier.  

I agree it's not ideal and can be a bit confusing, but my preference is we keeps things as they are and I'll continue to ask people every 3-4 months if they are ready to move to emeritus. Over time I think people with either re-engage or decide its time to move to emeritus.

marco-ippolito:
There is consensus that things are fine as they are so I'll close it

PR #1658, opened 2024-11-29
doc: use of OpenCollective and GitHub sponsors

mhdawson:


mhdawson:
@aduh95 I think you suggested we should document how the funds would be collected and managed as a first step in the discussions on funding.

This is my first cut at a simple process that we can expand on over time.

Nodejs-Open-Collective.md
```diff
@@ -0,0 +1,35 @@
...
+are shared with TSC members through 1password.
+
+## Disbursements - management
+
+Potential disbursements will be agreed through TSC discussion
+and before approval, documented in an PR in the
+[TSC repository](https://github.com/nodejs/TSC) which adds
+the intended use and funding envelope to this page. The disbursement
+is considered approved once TSC consensus in the PR has
+been reached.
```
aduh95, line 20: wdyt about adding a requirement for taking a vote if there's an objection? The idea would be to make it harder for TSC members to take controversial money decisions.
```suggestion
the intended use and funding envelope to this page. The disbursement
is considered approved if there's no objections from a project member
in 72 hours, and at least one approval from a TSC voting member.
Objections can be overridden with a TSC vote.
```

Nodejs-Open-Collective.md
```diff
@@ -0,0 +1,35 @@
...
+## Account details
+
+The account is manged by the TSC. The user id and password
+are shared with TSC members through 1password.
+
+## Disbursements - management
+
+Potential disbursements will be agreed through TSC discussion
+and before approval, documented in an PR in the
+[TSC repository](https://github.com/nodejs/TSC) which adds
```
aduh95, line 17: Should it be the admin repo?

Nodejs-Open-Collective.md
mhdawson: @aduh95 I assume you mean an objection from a non-TSC member as a vote would already be required if there was an objection from a TSC member.

I'd be ok with that.

mcollina: I'm ok with what is proposed, but I would propose something different. 

> Disbursements are approved in TSC meetings with quorum. In case of any objections, a TSC vote would be required.

mcollina (approved):
lgtm

Nodejs-GitHub-Sponsors.md
```diff
@@ -0,0 +1,34 @@
+# Node.js GitHub sponsors
```
mcollina, line 1: I think we should note somewhere that the funds received via GitHub Sponsors will be fiscally hosted by the OpenJS Foundation.

aduh95: I think that’s already the `Account details` section

Nodejs-Open-Collective.md
aduh95: I don’t know how consistent we are with this, my understanding is that nodejs/TSC would be for stuff internal to the TSC, so folks who are not TSC members can safely unsubscribe from that repo without missing any info relevant to them.

Nodejs-Open-Collective.md
mhdawson: @mcollina I think using quorum in a meeting does not align with making sure everybody has an opportunity to comment or object. I am happy to add the the clause about objections, but I think we should clarify in terms of "objections by who". Options could include:

* TSC members
* An members of the Node.js organization
* core collaborators (although I think this is not broad enough)
* Anybody (I presonally think this is too broad)


@mcollina, @aduh95 any other options than those and thoughts on which we should specify?

bensternthal:
In the CPC working session this morning myself @mhdawson Robin and Joe discussed the various fees one incurs with OpenCollective. We also reviewed the Linux Foundation (LF) General & Administrative (G&A) fee, which we wanted to highlight here for transparency.

If you choose to have OpenJS receive sponsorship dollars and manage contractor payments, the LF G&A fee will apply. This fee is structured as follows:

- 9% on the first $1M of funds
- 6% on all funds raised above $1M

Based on expected OpenJS’s annual revenue, we typically use an estimated 7.5% as the effective G&A fee for planning purposes.

Nodejs-Open-Collective.md
```diff
@@ -0,0 +1,36 @@
...
+the intended use and funding envelope to this page. The disbursement
+is considered approved if there's no objections from a project member
+in 72 hours, and at least one approval from a TSC voting member.
+Objections can be overridden with a TSC vote.
```
mhdawson, line 21:
```suggestion
is considered approved if there's no objections from a project member
in 72 hours, and consensus has been reached in the TSC. Objections can be
overridden with a TSC vote.
```

mhdawson: @aduh95 I merged your suggetion, but have added one as I think we need TSC consensus before funds are allocated. In my opinion having one TSC approval is not enough. We need TSC consensus which is a higher bar and will take a few weeks which I think is important in this case.

aduh95: SGTM

Nodejs-GitHub-Sponsors.md
```diff
@@ -0,0 +1,34 @@
...
+agreed to manage the funds that are received on behalf of the project.
+
+## Disbursements - management
+
+Potential disbursements will be agreed through TSC discussion
+and before approval, documented in a PR in the
+[TSC repository](https://github.com/nodejs/TSC) which adds
+the intended use and funding envelope to this page. The disbursement
+is considered approved once TSC consensus in the PR has
+been reached.
```
mhdawson, line 25:
```suggestion
 the intended use and funding envelope to this page. The disbursement
 is considered approved if there's no objections from a project member in 72 hours,
 and consensus has been reached in the TSC. Objections can be overridden with a TSC vote.
```

mhdawson:
@nodejs/tsc it would be great to get more people either adding their +1s or joining the conversation. Want to be sure we have consensus with the approach outlined.

RafaelGSS (approved):

mcollina (approved):
lgtm

mhdawson:
@aduh95 are you ok if these land here and then we decide in another PR to move all three to the admin repo ?

aduh95 (approved):

mhdawson:
@nodejs/tsc will mention in the meeting tomorrow and if there are not objections during the meeting or in the issue before then will assume its approved. 

In terms of the fees mentioned by @bensternthal seems like we are likely going to have some entity to act as a fiscal host and many of those charge fees which are +- a few percent from 7.5%.

anonrig (approved):

mhdawson:
Mentioned again in the meeting today, no objections so landing.

PR #1674, opened 2025-01-17
doc: consolidate funding source descriptions

mhdawson:
It was possible to have Open Collective funds be directed to the OpenJS foundation in a manner similar to the plan for GitHub sponsors. Update and consolidate the description of how funds will be managed.


mhdawson:
As a note both GitHub sponsors and Open Collective are enabled, but there is still work to define how we configure them, surface sponsors on the web page and add the userids/password for broader TSC access.

mcollina:
We can use a similar setup of Fastify https://github.com/sponsors/fastify. I would add a 10-20x increase in each tier.

jasnell (approved):

mhdawson:
@mcollina thanks for the suggestion to look at fastify as an example. I think that should be discussed separately from this PR which is just to bring our doc in line with what was actually set up.

mcollina (approved):
lgtm

PR #1775, opened 2025-08-03
Vote wrt James self-nomination to Moderation Team

aduh95:


mcollina:
My understanding is that this would be a public vote, right? I thought there was a specific ask for this to be secret.

votes/initiateNewVote/_EDIT_ME.yml
```diff
@@ -35,10 +35,23 @@ canShuffleCandidates: true
...
+  Refs: https://github.com/nodejs/TSC/issues/1773#issuecomment-3131028605
 
 # 6. Optionally, insert a brief introduction for the vote PR, in the markdown format.
 prBody: |
+  The ballots from this vote should not be communicated publicly, as suggested in
+  https://github.com/nodejs/TSC/issues/1773#issuecomment-3137209359.
+  After the vote has been opened for seven days and/or all TSC voting members have
+  sent an encrypted ballot, the secret parts should be communicated to the
+  Moderation Team which would be charged to communnicate the result, while keeping
+  the ballots secret.
```
aduh95, line 52: @mcollina the goal is for the vote to be anonymous

jasnell (approved):

votes/initiateNewVote/_EDIT_ME.yml
mcollina: The moderation team could see the individual votes then, no?

aduh95: That's correct, whoever we send the secret parts to would be able to read the ballots. I'm suggesting the moderation team as I believe they would be seen as a trusted entity – it doesn't have to be the whole Moderation Team, but IMO it's better if it's not a single individual.

legendecas (approved):

votes/initiateNewVote/_EDIT_ME.yml
jasnell: Agree with @aduh95 on this. I would trust anyone currently on the moderation team to view the results and trust that they'd keep it a secret. The one thing that I would want to make sure of is that if I do join the moderation team, those results are not kept in a location where I'd be able to see them after the fact.

RafaelGSS (approved):

anonrig (approved):

PR #1863, opened 2026-05-25
ci: cap GITHUB_TOKEN to contents: read

arpitjain099:
Pins the default `GITHUB_TOKEN` to `contents: read` at workflow level. No GitHub API writes from the workflow.

Post-CVE-2025-30066 hardening pattern. YAML validated locally.


aduh95 (approved):
FWIW it's already set as read-only in the repo settings, but having it set explicitly in the YAML is certainly fine and will cover forks as well

